PR #1, opened 2015-12-06
igmsys: IGMSystem, IGMSurvey, LLS, DLA

profxj:
First bits of code
Mainly pulled from linetools
tests and docs too


profxj:
Merging.

PR #3, opened 2015-12-08
fN: Model, tests and docs

profxj:
As titled.

See the ipython notebook for examples


pyigm/abssys/utils.py
```diff
@@ -23,6 +23,7 @@
 from linetools.analysis import absline as ltaa
 from linetools.isgm.abscomponent import AbsComponent
 
+
 def chk_components(components, chk_match=False, chk_A_none=False, toler=0.2*u.arcsec):
     """ Performs checks on a list of components
```
ntejos: I'm not convinced that this checks on Components should be done here, as opposed to do these in linetools. I thought we agreed that AbsSystem should also be part of linetools (at least for now).

profxj: You are right.  I had copied over the whole file to get some methods that
should be in pyigm but forgot to strip out the component ones.
Fixed.

profxj:
Note that the Travis failure is only due to an out-of-date astropy test (Plank15).

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+    mtype : str
+      Model type for the fN
+      'Hspline' -- Hermite monotonic spline
+      'Gamma' -- Gamma function [Following Inoue+14]
+    zmnx : tuple
+      Redshift range where this model applies (zmin,zmax)
+    pivots : array
+          log NHI values for the pivots
+    param : ndarray
+      Parameter array
```
mifumagalli: If this acts as documentation, let's be more specific on type of parameters?

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+          Recalculate the default model
+        use_mcmc : boolean, optional (False)
+          Use the MCMC chain to generate the model
+        write : boolean, optional (False)
+          Write out the model
+        """
+        if use_mcmc:
+            # MCMC Analysis (might put these on a website)
+            chain_file = (os.environ.get('DROPBOX_DIR')+
+                        'IGM/fN/MCMC/mcmc_spline_k13r13o13n12_8.fits.gz')
```
mifumagalli, line 63: Consider a data folder at top level to avoid too many hardcoded paths pointing to different locations? - edit: I see it at L70, so would use that and be clear in installation that people need to get extra stuff from a website.

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+            # Instantiate
+            fN_model = cls('Hspline', zmnx=(0.5, 3.0),
+                            pivots=np.array(fN_data['LGN']).flatten(),
+                            param=np.array(fN_data['FN']).flatten())
+        # Return
+        return fN_model
+
+    def __init__(self, mtype, zmnx=(0., 0.), pivots=None, param=None,
+                 zpivot=2.4, gamma=1.5):
+        self.mtype = mtype  # Should probably check the choice
```
mifumagalli: yes please - let's have warnings for values that are not allowed

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+        self.npivot = len(self.pivots)
+
+        # Param
+        if param is None:
+            self.param = np.zeros(self.npivot)
+        else:
+            self.param = param
+            # Init
+            if mtype == 'Hspline':
+                self.model = scii.PchipInterpolator(self.pivots, self.param,
```
mifumagalli: I guess this works, but for code readability  and future expansions, I would prefer to see the "if mtype =="  all at the same  level of indentation (and perhaps one after the other). I would therefore suggest a structure like this:

if pivots:
if param:
if mtype:

and then handle the special cases.

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+
+    def calculate_lox(self, z, NHI_min, NHI_max=None, neval=10000, cumul=False):
+        """ Calculate l(X) over an N_HI interval
+
+        Parameters
+        ----------
+        z : float
+          Redshift for evaluation
+        NHI_min : float
+          minimum NHI value
```
mifumagalli: We may wish to be more clear on log/no-log as applicable.

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+        for ii in range(nz): 
+            lX[ii] = np.sum(10.**(lgfNX[:, ii]+lgNHI)) * dlgN * np.log(10.)
+        if cumul == True:
+            if nz > 1:  #; Have not modified this yet
+                raise ValueError('fN.model: Not ready for this model type %s' % self.mtype)
+            cum_sum = np.cumsum(10.**(lgfNX[:, ii]+lgNHI)) * dlgN * np.log(10.)
+
+        # Infinity?
+        if infinity is True:
+            # This is risky...
```
mifumagalli, line 223: cut or keep?

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+        # Return
+        if nz == 1:
+            lX = lX[0]
+        if cumul == True:
+            return lX, cum_sum, lgNHI
+        else:
+            return lX
+
+    def calculate_rhoHI(self, z, NHI_mnx, neval=10000, cumul=False,
+        H0=70.*u.km/(u.s*u.Mpc)):
```
mifumagalli: Having this hardcoded seems dangerous if one wants to change cosmology

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+        -------
+
+        log_fNX : float, array, or 2D array
+          f(NHI,X)[z] values
+          Float if given one NHI,z value each. Otherwise 2D array
+          If 2D, it is [NHI,z] on the axes
+
+        """
+        # Tuple?
+        if isinstance(NHI, tuple): # All values packed into NHI parameter
```
mifumagalli: As a side comment, perhaps we should discourage overloading if not necessary

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+            beta = [item[1] for item in self.param[2:]] 
+            for kk in range(ncomp):
+                log_gN[:, kk] += (np.log10(Bi[kk]) + NHI*(-1 * beta[kk])
+                                + (-1. * 10.**(NHI-Nc) / np.log(10)))  # log10 [ exp(-NHI/Nc) ]
+            # f(z)
+            fz = np.zeros((lenz, 2))
+            # Loop on NHI
+            for kk in range(ncomp):
+                if kk == 0:  # LyaF
+                    zcuts = self.param[2][2:4]
```
mifumagalli: I guess this works ok for now, but using dictionary for param lists in the future may improve readability and avoid subtle bugs.

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+        except:
+            nz=1
+            z = np.array([z])
+
+        # Brute force (should be good to ~0.5%)
+        lgNHI = NHI_min + (NHI_max-NHI_min)*np.arange(neval)/(neval-1.)
+        dlgN = lgNHI[1]-lgNHI[0]
+
+        # Evaluate f(N,X)
+        lgfNX = self.evaluate(lgNHI, z)
```
mifumagalli: Are we ok in calling this without optional "cosmo"? H0 is used in L260, but dxdz in evaluate may use different H0? Risk of inconsistency?

pyigm/fN/fnmodel.py
```diff
@@ -0,0 +1,456 @@
...
+        # Cosmology
+        if cosmo is None:
+            cosmo = cosmology.core.FlatLambdaCDM(70., 0.3)
+
+        # Calculate teff
+        zval, teff_LL = pyteff.lyman_limit(self, zmin, zem, N_eval=neval, cosmo=cosmo)
+
+        # Find tau=1
+        imn = np.argmin(np.fabs(teff_LL-1.))
+        if np.fabs(teff_LL[imn]-1.) > 0.02:
```
mifumagalli: would not be better to construct interpolation function for this?

pyigm/fN/tau_eff.py
```diff
@@ -0,0 +1,307 @@
...
+
+    teff_LL = np.zeros(N_eval)
+
+    # dXdz
+    dXdz = pyigmu.cosm_xz(zval, cosmo=cosmo, flg_return=1)
+
+    # Evaluate f(N,X)
+    velo = (zval-zem)/(1+zem) * (const.c.cgs.value/1e5)  # Kludge for eval [km/s]
+
+    log_fnX = fN_model.evaluate(lgNval, zem, vel_array=velo)
```
mifumagalli: does this need a cosmo=cosmo?

pyigm/fN/tau_eff.py
```diff
@@ -0,0 +1,307 @@
...
+                  EW_spline=None, bval=24., fNz=False, cosmo=None, debug=False,
+                  cumul=None, verbose=False):
+    """ tau effective from HI Lyman series absorption
+
+    Parameters
+    ----------
+    ilambda : float
+        Observed wavelength (Ang)
+    zem : float
+        Emission redshift of the source [sets which Lyman lines are included]
```
mifumagalli, line 136: Needs to be completed (fN_model, EW_spline..)

pyigm/fN/tau_eff.py
```diff
@@ -0,0 +1,307 @@
...
+    bva : float
```
mifumagalli: bval

pyigm/fN/tau_eff.py
```diff
@@ -0,0 +1,307 @@
...
+            continue
+        # Cosmology
+        if fNz is False:
+            if cosmo not in locals():
+                cosmo = FlatLambdaCDM(H0=70, Om0=0.3) # Vanilla
+            #dxdz = (np.fabs(xigmu.cosm_xz(zeval-0.1, cosmo=cosmo)-
+            #            xigmu.cosm_xz(zeval+0.1,cosmo=cosmo)) / 0.2 )
+            #xdb.set_trace()
+            dxdz = pyigmu.cosm_xz(zeval,cosmo=cosmo,flg_return=1)
+        else: dxdz = 1. # Code is using f(N,z)
```
mifumagalli: nice option, but a bit inconsistent with what's above perhaps...  It would be nice to have this option at the fn level

pyigm/fN/tests/test_fnmodel.py
```diff
@@ -0,0 +1,60 @@
...
+    assert fN_I14.zmnx == (0., 10.)
+
+
+def test_default():
+    fN_default = FNModel.default_model()
+    #
+    assert fN_default.mtype == 'Hspline'
+    assert fN_default.zmnx == (0.5, 3)
+
+
```
mifumagalli, line 31: Should we add a small test for cosmology?

mifumagalli:
Done with first pass. See embedded comments.

pyigm/fN/fnmodel.py
profxj: Done.  Admittedly, these aren't meant to be modified
by the user.

I have now added explicit parameter documentation for the **init** command.

pyigm/fN/fnmodel.py
profxj, line 63: I think this is the only one and it is not intended to be used by anyone but me.

The others, e.g.             fN_file = (pyigm_path+'/data/fN/fN_spline_z24.fits.gz')
come with pyigm.

pyigm/fN/fnmodel.py
profxj: Done (for mtype).  The rest are rather open.

pyigm/fN/fnmodel.py
profxj: done

pyigm/fN/fnmodel.py
profxj: done (throughout).  We are always using log in the variable (I think).

pyigm/fN/fnmodel.py
profxj, line 223: keeping for now.

pyigm/fN/fnmodel.py
profxj: Agreed.  And I forget why I was forced to do this contortion.
Could try to undo it if you greatly prefer.

pyigm/fN/fnmodel.py
profxj: Good suggestion.  Have done so.

pyigm/fN/fnmodel.py
profxj: Good catch.  Fixed.
Now requesting an input cosmo (optional).

pyigm/fN/fnmodel.py
profxj: Yes, but I had a reason for this once. 
And I've forgotten it. ;)

Now interpolating.

pyigm/fN/tau_eff.py
profxj: Good catch.

pyigm/fN/tau_eff.py
profxj, line 136: Done (mainly).

pyigm/fN/tau_eff.py
profxj: fixed

pyigm/fN/tau_eff.py
profxj: Right.  Removed for now.

pyigm/fN/tests/test_fnmodel.py
profxj, line 31: I plan to add some additional tests once I ingest
some f(N) constraints.  Will do that once the FNModel
syntax settles.

mifumagalli:
1) I still see a couple of calls to .evaluate without cosmo in tau_eff.py
2) in fnmodel.py: param is defined as dictionary but not used as such (L160+)
3) branch has conflict?

Once issues 1-3 resolved, happy to merge.

profxj:
1) Right you were.  Fixed
2) Good catch.  Fixed
3) I'll check for that.  Fixed.

mifumagalli:
1) How about .evaluate in L231 for fnmodel.py

profxj:
Got it.  And I edited travis so that the tests might pass.

Issue #7, opened 2015-12-21
Metallicity PDF class


astronomeara:
Would be nice to quote central value and confidence levels

profxj:
on it

profxj:
done

PR #8, opened 2015-12-21
MetallicityPDF: confidence intervals

profxj:
Introduces the MetallicityPDF class
Addresses Issue #7
test and docs and notebook too


profxj:
I got myself confused on PR's.  I thought I had issued
one for MetallicityPDF already.  This is now the main one.

@mifumagalli , please inspect it with JMO.

pyigm/metallicity/pdf.py
```diff
@@ -0,0 +1,154 @@
...
+        Parameters
+        ----------
+        ZH : ndarray
+          Values of ZH (log10, typically but not required) where the PDF has been evaluated
+        pdf_ZH : ndarray
+          PDF of the metallicity
+        parent : object, optional
+          Parent of the metalliicty PDF, e.g. an AbsSystem
+        dZH : float or ndarray
+          Width of the ZH 'bins' (log 10)
```
mifumagalli: Note that must be regularly spaced, given line 49

pyigm/metallicity/pdf.py
```diff
@@ -0,0 +1,154 @@
...
+        frac = (1.- cl) / 2.
+        ZH_min = float(f(frac))
+        ZH_max = float(f(1-frac))
+        # Return
+        return ZH_min, ZH_max
+
+    def normalize(self):
+        """ Normalize the PDF
+
+        First checks to see if it is normalized
```
mifumagalli: Do we want to check? - probably a good idea for those PDFs with some roundoff errors where the integral is not ==1

mifumagalli:
Quite happy with it - modulo query above. When fixed, happy to merge if JMO approves.

astronomeara:
go for it.  I’m still catching up on the process, but defer to your judgement

pyigm/metallicity/pdf.py
profxj: I was going to check and then decided not to (but left
the comment in).  I think it best not to check but just do.

profxj:
Removed the offensive comment.
Pushed.

PR #10, opened 2015-12-29
CGM: Classes for CGM analysis + COS-Halos

profxj:
Classes for CGM analysis
 includes code (but limited docs, tests) for COS-Halos
docs for main classes
simple tests
notebook


docs/cgm.rst
```diff
@@ -0,0 +1,120 @@
...
+=========
+
+This class enables an absorption-line analysis of a CGM.  This has been
+the primary approach to CGM analysis to date.  This class requires
+both a Galaxy object and :doc:`igmsys` object for instantiation.
+
+=========   ================= ============================================
+Attribute   Type              Description
+=========   ================= ============================================
+galaxy      Galaxy object     Describes the galaxy hosting the CGM
```
jnburchett, line 61: Can the CGMAbsSys galaxy be different from the CGM galaxy?  This could be useful if we are investigating multiple satellites having their own CGMAbsSys objects but composing a CGM object based a massive central?  Thoughts how we would handle this scenario?

docs/cgm.rst
```diff
@@ -0,0 +1,120 @@
...
++++++++
+
+This is overloaded to return an array of properties from one of the
+internal sets of objects in the survey.  The order of attribution is
+CGMAbsSys objects, Galaxy objects, and then IGMSystem objects. ::
+
+   rho_array = cgmsurvey.rho # Grabs rho from CGMAbsSys objects
+   z_array = cgmsurvey.z  # Grabs galaxy redshifts
+   coord = cgmsurvey.coord # Grabs galaxy coordinates
+
```
jnburchett, line 120: Lots of opportunities (should keep in mind as Galaxy class evolves):  
cgmsurvey.mass   # return stellar mass distribution
cgmsurvey.ion_detections('O VI')   # return column densities of given ion for the CGM objects
cgmsurvey.sSFR    # return sSFR distribution

docs/cgm.rst
```diff
@@ -0,0 +1,120 @@
...
+============
+
+This class organizes a survey of CGMAbsSys objects, e.g. COS-Halos.
+
+=========   ================= ============================================
+Attribute   Type              Description
+=========   ================= ============================================
+survey      str               Name of the survey, e.g. COS-Halos
+ref         str               References for the survey
+cgm_abs     list              list of CGMAbsSys objects
```
jnburchett, line 90: To continue the hierarchy, shouldn't the CGMAbsSurvey comprise CGM objects instead of CGMAbsSys?

pyigm/cgm/cgmsurvey.py
```diff
@@ -0,0 +1,90 @@
...
+      Survey name
+    ref : str, optional
+      Reference(s)
+    """
+    # Initialize with a .dat file
+    def __init__(self, survey='', ref=''):
+
+        # Name of survey
+        self.survey = survey
+        self.ref = ref
```
jnburchett, line 32: Could the survey be a list of journal, vol, pg?  Then, reference info could be populated upon instantiation or at some other time.

pyigm/cgm/cgm.py
```diff
@@ -0,0 +1,107 @@
...
+    ----------
+    rho : Quantity
+      Impact parameter (u.kpc)
+    """
+
+    # Initialize
+    def __init__(self, galaxy, igm_sys, cosmo=None):
+        # Checks
+        if not isinstance(galaxy, Galaxy):
+            raise IOError('CGMAbsSys instantiated with a Galaxy')
```
jnburchett, line 111: "..._must_ be instantiated with a Galaxy"

pyigm/cgm/cgm.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Calculate rho
+        if cosmo is None:
+            from astropy.cosmology import WMAP9 as cosmo
+            warnings.warn('cgm.CGMAbsSys: Using WMAP9 cosmology')
+            self.cosmo = cosmo
+        else:
+            self.cosmo = cosmo
+        ang_sep = self.igm_sys.coord.separation(self.galaxy.coord).to('arcmin')
+        kpc_amin = cosmo.kpc_comoving_per_arcmin( self.galaxy.z)  # kpc per arcmin
```
jnburchett: For very low-z, do LG, Virgo, etc., velocity correction?

pyigm/cgm/cos_halos.py
```diff
@@ -0,0 +1,423 @@
...
+        self.cgm_abs[mm].igm_sys.NHI = dat_tab[
+            (dat_tab['Z']==1)&(dat_tab['ion']==1)]['logN'][0]
+        self.cgm_abs[mm].igm_sys.flag_NHI = dat_tab[
+            (dat_tab['Z']==1)&(dat_tab['ion']==1)]['flag_N'][0]
+
+    # Load from mega structure
+    def load_mega(self,data_file=None, cosh_dct=None, test=False, **kwargs):
+        """ Load the data for COS-Halos
+
+        Paramaeters
```
jnburchett: 'Parameters'

pyigm/cgm/cos_halos.py
```diff
@@ -0,0 +1,423 @@
...
+        for fil in cos_files:
+            self.load_single(fil, **kwargs)
+
+    
+    ########################## ##########################
+    def load_abskin(self,flg=1,kin_file=None,kin_init_file=None):
+        """ Load the absorption-line kinematic data for COS-Halos (or COS-Dwarfs)
+        Calculate from scratch if needed
+
+        Paramaeters
```
jnburchett: 'Parameters'

pyigm/cgm/cos_halos.py
```diff
@@ -0,0 +1,423 @@
...
+        data_file : string
+          Name of data file
+        pckl_fil : string
+          Name of file for pickling
+        """
+        # Loop
+        if test is True:
+            cos_files = glob.glob(self.fits_path+'/J091*.fits.gz')  # For testing
+        else:
+            cos_files = glob.glob(self.fits_path+'/J*.fits.gz')
```
jnburchett, line 262: If I'm following this correctly, those fits.gz files should be within something like "~/Dropbox/COS-Halos/Summary/", but I don't see them in my copy of the Dropbox?  For the COS-Dwarfs data, I converted the IDL .sav files to .fits, and all the data were in 'cosmetals_megastructure.new.sav'.   Is there a need to loop through the structures for each COS-Halos galaxy?

docs/cgm.rst
profxj, line 61: Yes, these can be distinct although I hadn't thought
about it previously.  But the CGMAbsSys objects
form a list in CGM and there is no requirement on 
matching galaxy.

docs/cgm.rst
profxj, line 120: Agreed.  We can/should add to the documentation as we go.

docs/cgm.rst
profxj, line 90: No, the intent would be to have a CGMSurvey object.  
But that doesn't exist yet nor do I think that I have an
application for it yet.

pyigm/cgm/cgmsurvey.py
profxj, line 32: Neither survey nor ref need be provided.

pyigm/cgm/cgm.py
profxj, line 111: Edited.

pyigm/cgm/cgm.py
profxj: Absolutely.  As that is not yet implemented, at
what redshift do you recommend that we throw
a warning or error?

pyigm/cgm/cos_halos.py
profxj: Fixed

pyigm/cgm/cos_halos.py
profxj: Fixed.

pyigm/cgm/cos_halos.py
profxj, line 262: COS-Halos-Data/Summary/

You shouldn't have to do anything.

docs/cgm.rst
jnburchett, line 61: I think the generality allows for the scenario I describe then!

docs/cgm.rst
jnburchett, line 90: I guess I'm not clear on the utility of having both then?

profxj, line 90: COS-Halos is and will likely always be a CGMAbsSurvey.
Not sure it will become a CGMSurvey.  Maybe.

pyigm/cgm/cgm.py
jnburchett: Beyond z=0.05, it's less than a 6% effect.  At z<0.05, the relative effect rises sharply, so maybe z=0.05?

profxj: It will now raise an Error.

Would appreciate your help adding code to address this.

profxj:
Merging (at long last!)

PR #14, opened 2016-01-22
Continuum:  JFH's full SED model

profxj:
Implements JFH's model SED for a QSO
  Uses Lusso, Richard's, Vanden Berk, etc.
  Adds a Notebook
  test
  no other docs


profxj:
Did you even look at it! ;)

X

---

J. Xavier Prochaska
UCO/Lick Observatory
1156 High St.
UC Santa Cruz
Santa Cruz, CA 95064

http://www.ucolick.org/~xavier/
http://imps.ucolick.org/
http://www.qpqsurvey.org/
http://k1dm3.ucolick.org/
http://www.ucolick.org/~xavier/IDL/index.html
https://github.com/profxj/xastropy

831-459-2135 (Direct)
831-459-2991 (UCO/Lick Main)
831-459-5244 (Fax)

astronomeara:
In fact, I did!

astronomeara:
I think at the end of the day, we may just have some sort of systematic error from the lack of ‘real’ EUV tilt information.  That error will grow as a function of wavelength shortward of 912 (i.e., it gets bigger the farther down in rest wavelength you go without first hitting a tau of 2ish LLS).  One could probably explicitly calculate the probability for some value of a lambda < 912 tilt and an f(N) how many times one incorrectly fits a tau ~ 0.5 (or 1?) lls simply because they got the wrong tilt and saw a shelf that actually wasn’t there.

One could further postulate that I’m just over-thinking it…..

J

PR #16, opened 2016-01-24
MetallicityPDF: MCMC routines

profxj:
Compliments of Fumagalli
  see Fumagalli+16 for more

No tests yet
Limited docs


pyigm/metallicity/mcmc.py
```diff
@@ -0,0 +1,887 @@
...
+
+        Returns
+        -------
+
+        """
+
+        print('Writing outputs...')
+
+        #pickle the results to disk
+        wout=open(self.outsave+'/'+self.info['name']+'_emcee.pkl','w')
```
mifumagalli: I would choose a format type and stick to it - I started with pickle, but I agree there is better staff around. hdf5? :)

pyigm/metallicity/mcmc.py
```diff
@@ -0,0 +1,887 @@
...
+    Returns
+    -------
+
+    """
+
+    print('Getting started...')
+
+    #Do not run with less than 100 samples
+    nsamp=np.max([nsamp,100])
+
```
mifumagalli, line 880: Given the call below in pyigm_mtlmcmc.py, we should perhaps add a floor to nwalkers as well. Something like    nwalkers=np.max([nwalkers,400])

pyigm/scripts/pyigm_mtlmcmc.py
```diff
@@ -0,0 +1,79 @@
...
+                     'z':data['zabs'][sets[ii]],'errz':data['ezabs'][sets[ii]],'name':sy.strip()}
+        else:
+            #append ion value to list
+            observ.append((data['ion'][sets[ii]],data['logn'][sets[ii]],data['elogn'][sets[ii]],data['flag'][sets[ii]]))
+
+    #run the mcmc
+    print 'Ready to run mcmc for {}'.format(sy)
+
+    #pick optimised values for 12 processors - cosma (proc*NN walkers, proc*YY samples)
+    mcmc=mcmc_ions(observ,obsinfo,args.grid,nwalkers=(args.proc*80),nsamp=(args.proc*40),
```
mifumagalli: nwalkers=(args.proc_80),nsamp=(args.proc_40) was optimised for runs on Durham cluster. Given we are now writing this for single systems on shared memory machine, I would make nwalkers and nsamp optional paramaters that one can influence from command line.

pyigm/scripts/pyigm_mtlmcmc.py
```diff
@@ -0,0 +1,79 @@
...
+
+def main(args=None):
+    import argparse
+    #get the call
+    parser = argparse.ArgumentParser(description='Running grid on shared memory system')
+    parser.add_argument('sy', type=str, help='Name of the System to analyze')
+    parser.add_argument('fileinput')
+    parser.add_argument('outsave')
+    parser.add_argument('grid')
+    parser.add_argument('proc', type=int)
```
mifumagalli: As said above, I would make nproc and nsamp accessible to users.

mifumagalli:
I will test outputs soon. Meanwhile, please consider suggestions I made. Thanks.

pyigm/metallicity/mcmc.py
profxj: I'm willing to do hd5.  I'm not keen on Python specific files, just
like I never liked using IDL save files.  Am happy to defer to
your wisdom here.  I presume these may get very large if 
saved as ASCII (i.e. JSON).  So a binary format makes sense.
But it might be that a mix of the two is best, i.e. a JSON file for
the key parameters and an HDF5 file for the PDF output.  Thoughts?

pyigm/metallicity/mcmc.py
profxj, line 880: Done.

pyigm/scripts/pyigm_mtlmcmc.py
profxj: Ok, have put them in as optional.  I have also replaced number of
processors with nthread.

profxj:
Edits made.

pyigm/metallicity/mcmc.py
mifumagalli: I would avoid a mix. If we want binary, then hdf5 is probably the best as it is portable and can store a mix of data and attributes (comments).

profxj: hd5 it shall be

X

> On Jan 27, 2016, at 10:04 AM, Michele Fumagalli notifications@github.com wrote:
> 
> In pyigm/metallicity/mcmc.py https://github.com/pyigm/pyigm/pull/16#discussion_r51023138:
> 
> > -    def plotinfo(self, sampler):
> > -        """ Function that makes some useful plots of stuff
> >   +
> > -        Parameters
> > -        ----------
> > -        sampler
> >   +
> > -        Returns
> > -        -------
> >   +
> > -        """
> >   +
> > -        print('Writing outputs...')
> >   +
> > -        #pickle the results to disk
> > -        wout=open(self.outsave+'/'+self.info['name']+'_emcee.pkl','w')
> >   I would avoid a mix. If we want binary, then hdf5 is probably the best as it is portable and can store a mix of data and attributes (comments).
> 
> —
> Reply to this email directly or view it on GitHub https://github.com/pyigm/pyigm/pull/16/files#r51023138.
> 
> !DSPAM:826,56a906cd213683159824774!

---

J. Xavier Prochaska
UCO/Lick Observatory
1156 High St.
UC Santa Cruz
Santa Cruz, CA 95064

http://www.ucolick.org/~xavier/
http://imps.ucolick.org/
http://www.qpqsurvey.org/
http://k1dm3.ucolick.org/
http://www.ucolick.org/~xavier/IDL/index.html
https://github.com/pyigm/pyigm
https://github.com/linetools/linetools

831-459-2135 (Direct)
831-459-2991 (UCO/Lick Main)
831-459-5244 (Fax)

profxj: Implemented.

mifumagalli:
Ok to merge, but still running some tests.

PR #4, opened 2015-12-13
field:  The Beginnings

profxj:
Class with a few methods for an IGM-Galaxy field

docs
one simple test
notebook

PR #17, opened 2016-01-29
DLA Surveys

profxj:
More from the literature, compliments of Ruben

Peroux03
G09
XQ-100


mneeleman:
Four small comments:
1) Two wrong references: P03 should be Peroux et al 2003 MNRAS 346 1103 and G09 should be A&A (not ApJ).
2) the end redshifts are not from these papers, but are from the 5000 km/s cut-off used by Ruben. Not a problem, but I feel that should be mentioned somewhere.
3) The IUE data from the P03 sample has been removed in this data set. Again, this probably should be explicitly mentioned.
4) Should we add the HST sample?

profxj:
Addressed Marcel's comments (except #4).

Merging despite test failure..

PR #20, opened 2016-03-30
Mock Forest spectra

profxj:
Generates mock Lya forest spectra from
f(N), wave, and zem of the source.

Added to fN Notebook and docs
Test


profxj:
One test will pass eventually..

PR #23, opened 2016-06-15
IGMGuesses GUI

profxj:
Moved from xastropy to here.
Added a script to run it and modified setup.py accordingly.

Added a file in docs for @ntejos and @jnburchett to contribute to.
No tests nor Notebook

### DIFF
--- a/docs/igmguesses.rst
+++ b/docs/igmguesses.rst
@@ -23,8 +23,8 @@ these initial guesses should be then used for subsequent Voigt profile
 fitting. Still, IGMGuesses does provide an easy handling of blends,
 which can become a problem for high-redshift or high-S/N data.
 
-Usage
-=====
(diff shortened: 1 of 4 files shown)


ntejos:
I suggest to merge it and test it from there... (?)

profxj:
I can do that if you prefer, but you can just as easy get on this branch and
test it.  I'm hoping that one of you has a moment to add a little documentation
before we merge this.

ntejos:
I see; let's do that then. Will test the branch on my laptop and add some documentation before merging ;)

profxj:
Note that the failing tests are addressed in the reion branch 
and may be ignored here.

profxj:
@ntejos please merge when you feel happy to do so.
I won't be editing it further for now.

PR #27, opened 2016-06-24
LLS dict

profxj:
a few things for LLS


profxj:
Failing for Python 3.5 but I think that is bogus.
Merging.

PR #28, opened 2016-06-24
Lls

profxj:
A few more bits and pieces


profxj:
merging (still failing 3.5 tests)

PR #29, opened 2016-06-26
CGM notebook and a bit more

profxj:
As titled


profxj:
Failling 3.5 but otherwise fine

profxj:
Merging

PR #32, opened 2016-07-06
COS Dwarfs

profxj:
COS Dwarfs and a bit more


pyigm/cgm/cgm.py
```diff
@@ -117,7 +135,7 @@ def __init__(self, galaxy, igm_sys, cosmo=None, name=None, rho=None, PA=None, an
         self.igm_sys = igm_sys
 
         # Raise error for redshifts not in the Hubble flow
-        if galaxy.z < 0.05:
+        if (galaxy.z < 0.05) and chk_lowz:
```
jnburchett, line 139: Ah yes, a call to arms!  Got it loud and clear.

profxj:
merging.  Failing only on Planck15 import;  not sure why

PR #40, opened 2016-07-23
fN MCMC code

profxj:
ported from xastropy

no tests or docs

used on P14 data.

Am about to port to pymc3


profxj:
Failing dndx_rvir which is addressed in PR#39

profxj:
merging

PR #42, opened 2016-07-25
More on igmguesses

ntejos:
IGMGuesses: component widget selects the same component as selected using 'S' in the main velocity plots. This helps to better find your way between the three main widgets (velocity, list, and fiddle).


profxj:
Merge if you have tested it yourself.

PR #44, opened 2016-07-27
Igmguesses: enhancements

ntejos:
- Compatibility of the "colorful" option with .json files saved with previous versions of IGMGuesses.
- Towards a clever component grouping depending on overlap in wobs space (still in develop).

### DIFF
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -484,8 +484,6 @@ P         : toggle on/off "colorful" display, where components of different
             f.write(unicode(json.dumps(gd_dict, sort_keys=True, indent=4,
                                        separators=(',', ': '))))
 
-
-
     # Write + Quit
     def write_quit(self):
         self.write_out()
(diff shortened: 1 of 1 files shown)


pyigm/guis/igmguesses.py
```diff
@@ -1003,6 +1015,11 @@ def on_key(self,event):
         if event.key == 'P':  # Toggle colorful mode
             self.flag_colorful = not self.flag_colorful
 
+        # print component info
+        if event.key == '@':  #  for develop; this will be deleted in future.
```
profxj, line 1019: I'm unclear.  Is this to be used?  If not, why is it here?

Other than this issue, all looks fine.  Merge when ready

ntejos, line 1019: This is only intended to be use for developers, because `blending_info` is taking too long to compute for sightlines with more than, say, 10 components which I find annoying. Still, for the patient this provides useful information so I want to keep it for now. My goal is to find out a better way to obtain this information that hopefully increases speed.

My thoughts are toward adding an extra attribute to components, which stores the relation of it w/r to all the rest (in the sense of overlapping in wobs space) in a given list. But may need to work on issue https://github.com/linetools/linetools/issues/260 before proceeding further.

profxj:
Merging.

PR #50, opened 2016-08-26
IGMGuesses: bug solved when writing the full .joebvp input file from …

ntejos:
….json file.


profxj:
Looks fine.  Merge when tests pass.

PR #54, opened 2016-09-15
NH calculation when NHI is out of bounds

profxj:
As titled.

Tested on COS-Halos by JXP

### DIFF
--- a/pyigm/metallicity/utils.py
+++ b/pyigm/metallicity/utils.py
@@ -10,7 +10,7 @@ import pdb
 
 
 def calc_logNH(hdf_file, modl=None, sys_error=0., NH_interpol=None, init_interpol=False,
-                test=False, ret_flags=['NH'], debug=False):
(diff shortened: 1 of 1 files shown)


mifumagalli (commented):
What's the logic behind this? It does not seem to be a self consistent treatment.

profxj:
There are several COS-Halos systems with NHI < 10^15 which falls
off the standard Cloudy grid.  To perform the analysis on these, which
are optically thin, I reset their NHI to the edge of the grid (e.g. 1e15) 
perform the analysis and then shift back down.  That is the intent here
and I think I coded it that way.

mifumagalli (changes requested):
Ok, can I suggest something like this around L80

if(min_NHI > 16):
    warnings.warn('Warning: getting close to optically thick regime')

Then, merge.

profxj:
Good suggestion.  Added.

Merging.

PR #56, opened 2016-09-18
dndX

profxj:
Modifications to the dN/dX calculation

### DIFF
--- a/pyigm/cgm/analysis.py
+++ b/pyigm/cgm/analysis.py
@@ -22,7 +22,7 @@ from astropy import units as u
 from astropy import constants as const
 from astropy.cosmology import Planck15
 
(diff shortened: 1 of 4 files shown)


pyigm/cgm/analysis.py
```diff
@@ -53,11 +64,18 @@ def dndx_rvir(Lrng=(0.001, 10), beta=0.2, rvir_Lstar=250.*u.kpc,
     phi_str_cgs = (phi_str_pref * 1e-2 * hubb**3) / u.Mpc**3
```
jnburchett, line 64: Line 42 or 47 should contain a note that the phi_str needed is actually Phi\* times 100.

jnburchett (approved):
Everything in this PR looks good...one clarification in docstring requested.

profxj:
merging

PR #62, opened 2016-10-06
COS-Halos dataset

profxj:
Releases v1.1 of the COS-Halos dataset.

Does not include spectra (see igmspec, v02).

Does include a little bit of docs and the previous
Notebook.


profxj:
Merging

PR #66, opened 2016-10-21
COS-Halos v1.1 and NHI for P11

profxj:
As titled.  Added ZH, and sig_ZH for COS-Halos

and NHI for P11

Notebooks updated too


jnburchett (approved):
Whooo-hooo!

PR #73, opened 2016-11-25
Igm guesses

ntejos:
Compatibility with new linetools' absline.z attribute
Use ltu.give_dv() when possible.
Hidden '9' keystroke to test H2 molecules (in development)


pyigm/guis/igmguesses.py
```diff
@@ -1208,7 +1220,8 @@ def on_draw(self, replot=True, in_wrest=None, rescale=True, fig_clear=False):
                 # Velocity
                 wvobs = (1 + self.z) * wrest
                 wvmnx = wvobs * (1 + np.array(self.psdict['x_minmax']) / c_kms)
-                velo = (self.spec.wavelength/wvobs - 1.) * c_kms * u.km/u.s
+                # velo = (self.spec.wavelength/wvobs - 1.) * c_kms * u.km/u.s
+                velo = ltu.give_dv(self.spec.wavelength/wrest - 1., self.z)
```
profxj, line 1226: Can't say that I love the name   give_dv()
but we are probably stuck with it for now..

pyigm/guis/igmguesses.py
```diff
@@ -1236,22 +1249,29 @@ def on_draw(self, replot=True, in_wrest=None, rescale=True, fig_clear=False):
...
                                     rotation=90.,ha='center',va='center')
 
                         if (self.flag_plotmodel) and (self.flag_colorful):
                             # plotting absline with color better done in wobs -> velo space
-                            dvmin = c_kms * ((line_wvobs_lims[imt][0] - line_wvobs[imt]) / wvobs)
-                            dvmax = c_kms * ((line_wvobs_lims[imt][1] - line_wvobs[imt]) / wvobs)
-                            cond = (velo.value >= v + dvmin) & (velo.value <= v + dvmax)
+                            # dvmin = c_kms * ((line_wvobs_lims[imt][0] - line_wvobs[imt]) / wvobs)
+                            # dvmax = c_kms * ((line_wvobs_lims[imt][1] - line_wvobs[imt]) / wvobs)
+                            vmin = ltu.give_dv(line_wvobs_lims[imt][0]/wrest - 1., self.z)
```
profxj: Might be able to do with 1 call to give_dv(), i.e.

vmin, vmax = ltu.give_dv(....)

pyigm/guis/igmguesses.py
```diff
@@ -1771,7 +1792,10 @@ def from_igmguesses_to_joebvp(infile, outfile):
     # Components
     comp_list = []
     for ii, key in enumerate(igmg_dict['cmps'].keys()):
-        comp = AbsComponent.from_dict(igmg_dict['cmps'][key], chk_sep=False, chk_data=False, chk_vel=True)
+        # QtCore.pyqtRemoveInputHook()
+        # pdb.set_trace()
+        # QtCore.pyqtRestoreInputHook()
+        comp = AbsComponent.from_dict(igmg_dict['cmps'][key], chk_sep=False, chk_data=False, chk_vel=False)
```
profxj, line 1798: Maybe explain why to have chk_vel=False here.  Unless it is obvious (it isn't to me).

profxj (changes requested):

pyigm/guis/igmguesses.py
ntejos: Not necessarily, if you have a better name in mind please suggest and I do the refactor. I remember there is something like `v_from_z` and `z_from_v` which at the moment are slightly different than `give_dz` or `give_dv` because they do not make explicit where is the reference system.

pyigm/guis/igmguesses.py
ntejos: Indeed. thanks!

pyigm/guis/igmguesses.py
ntejos: For some reason this check is removing some of my lines. It may be an issue with backwards compatibility from LineLimits. It seems I have too many limits in some of my .json files that need to be clean at some point in the future. For this particular function, we can assume that the input .json file was generated by igmguesses, in which case all lines within a given components are good.

profxj: ok, sounds fine.

pyigm/guis/igmguesses.py
profxj: Yeah, we should definitely avoid having both of these sets of methods.

Please do go for a refactor.  I suspect v_from_z is used in quite a few places
so probably best to edit that one.  Will raise as an Issue in linetools

PR #79, opened 2016-12-17
Build IGM systems in IGMSightline

profxj:
As titled

### DIFF
--- a/pyigm/igm/igmsightline.py
+++ b/pyigm/igm/igmsightline.py
@@ -10,6 +10,7 @@ from astropy.coordinates import SkyCoord
 
 from linetools.isgm.abssightline import AbsSightline
 from linetools.isgm.abssystem import add_comps_from_dict
+from linetools.isgm.utils import build_systems_from_components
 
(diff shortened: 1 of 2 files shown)


jnburchett (approved):
Looks great. Merging

pyigm/igm/igmsightline.py
```diff
@@ -10,6 +10,7 @@
 
 from linetools.isgm.abssightline import AbsSightline
```
jnburchett, line 11: Tests are failing here?

jnburchett (approved):
Oops, just noticed that a couple of the tests are failing on line 11.  No idea why, though, because this import is totally valid and works on my end.  I'm fine with merging if we are okay with the test failing.

pyigm/igm/igmsightline.py
profxj: This isn't showing on Travis.  I suspect you need to update linetools

profxj:
Merging

PR #84, opened 2017-01-06
Call xspec from inside igmguesses

profxj:
Adds keystroke "&" to enable the igmguesses user to launch
a new widget (a modified xspec).

It also converts the list of componetns into AbsSystems
and each of these is plotted (with unique color coding).

No tests or docs (aside from top doc string).

It may make sense to push the new Dialog into linetools.
Don't merge until we decide on that.

### DIFF
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -147,7 +147,7 @@ P         : toggle on/off "colorful" display, where components of different
 E         : toggle displaying/hiding the external absorption model
 %         : guess a transition and redshift for a given feature at
             the cursor's position
-&         : Launch an ExamineSpecWidget to view the spectrum and lines
+&         : Launch an ExamineSpecWidget to view the spectrum and any absorption systems
 ?         : print this help message
 """
 
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -1704,13 +1704,17 @@ def comp_init_attrib(comp):
 def sync_comp_lines(comp, only_lims=False):
     """Synchronize attributes of the lines and updates
     """
+    if only_lims:
+        # Need to update zlim *not* vlim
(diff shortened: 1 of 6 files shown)


ntejos:
What would be the main idea of having xspec loaded within IGMGuesses?

profxj:
1. View the full spectrum vs. wavelength
1. Inspect which lines are still not identified (and which are)

pyigm/guis/igmguesses.py
```diff
@@ -1836,4 +1854,69 @@ def from_igmguesses_to_joebvp(infile, outfile):
...
+            z = event.xdata/wrest.value - 1.
+            self.pltline_widg.llist['z'] = z
+
+            self.pltline_widg.zbox.setText('{:.5f}'.format(self.pltline_widg.llist['z']))
+
+            # Draw
+            self.spec_widg.on_draw()
+
+    def quit(self):
+        self.close()
```
ntejos: As discussed, may be better to move this ExSpecDialog to linetools.

pyigm/guis/igmguesses.py
```diff
@@ -1686,13 +1704,17 @@ def comp_init_attrib(comp):
 def sync_comp_lines(comp, only_lims=False):
     """Synchronize attributes of the lines and updates
     """
+    if only_lims:
+        # Need to update zlim *not* vlim
+        comp_vlim_mks = comp.vlim.to('km/s').value
+        zlim_comp = comp.zcomp + (1 + comp.zcomp) * (comp_vlim_mks / c_kms)
```
ntejos: Use `ltu.z_from_dv()` instead

ntejos (changes requested):
Looks good, please merge after addressing the comment.

pyigm/guis/igmguesses.py
profxj: replaced as requested.

profxj:
Merging

PR #88, opened 2017-02-06
Peculiar velocity corrections

jnburchett:
Functions to correct for peculiar velocities when calculating distances from redshifts
Includes test

### DIFF
--- a/pyigm/cgm/cgm.py
+++ b/pyigm/cgm/cgm.py
@@ -107,7 +107,7 @@ class CGMAbsSys(object):
         return slf
 
     def __init__(self, galaxy, igm_sys, cosmo=None, name=None, rho=None, PA=None,
-                 ang_sep=None, chk_lowz=True, **kwargs):
+                 ang_sep=None, correct_lowz=True, **kwargs):
         """
         Parameters
         ----------
@@ -118,8 +118,9 @@ class CGMAbsSys(object):
         rho
         PA
         ang_sep
-        chk_lowz : bool, optional
-          Demand that z>0.05  [for impact parameter calculation]
+        correct_lowz : bool, optional
+          If galaxy z < 0.05, correct for peculiar velocies in impact parameter
+          calculation
 
         Returns
         -------
@@ -136,10 +137,6 @@ class CGMAbsSys(object):
             raise IOError('CGMAbsSys instantiated with an IGMSystem')
         self.igm_sys = igm_sys
 
-        # Raise error for redshifts not in the Hubble flow
-        #if (galaxy.z < 0.05) and chk_lowz:
-        #    raise NotImplementedError("Not prepared for such low redshift.  Need to implement corrections.")
-
         # Calculate rho
         if cosmo is None:
             warnings.warn('cgm.CGMAbsSys: Using WMAP9 cosmology')
@@ -149,7 +146,7 @@ class CGMAbsSys(object):
 
         # Impact parameter and PA
         if rho is None:
-            rho, iang = calc_rho(galaxy, igm_sys, self.cosmo, ang_sep=ang_sep, chk_lowz=chk_lowz)
+            rho, iang = calc_rho(galaxy, igm_sys, self.cosmo, ang_sep=ang_sep, correct_lowz=correct_lowz)
             if ang_sep is None:
                 ang_sep = iang
         self.rho = rho
--- a/pyigm/cgm/cgm.py
+++ b/pyigm/cgm/cgm.py
@@ -137,8 +137,8 @@ class CGMAbsSys(object):
         self.igm_sys = igm_sys
 
         # Raise error for redshifts not in the Hubble flow
-        if (galaxy.z < 0.05) and chk_lowz:
-            raise NotImplementedError("Not prepared for such low redshift.  Need to implement corrections.")
+        #if (galaxy.z < 0.05) and chk_lowz:
+        #    raise NotImplementedError("Not prepared for such low redshift.  Need to implement corrections.")
 
         # Calculate rho
         if cosmo is None:
@@ -149,7 +149,7 @@ class CGMAbsSys(object):
 
         # Impact parameter and PA
         if rho is None:
-            rho, iang = calc_rho(galaxy, igm_sys, self.cosmo, ang_sep=ang_sep)
+            rho, iang = calc_rho(galaxy, igm_sys, self.cosmo, ang_sep=ang_sep, chk_lowz=chk_lowz)
             if ang_sep is None:
                 ang_sep = iang
         self.rho = rho
--- a/pyigm/cgm/tests/test_init_cgm.py
+++ b/pyigm/cgm/tests/test_init_cgm.py
@@ -70,3 +70,15 @@ def test_cgm_from_igmsystems():
     cgm_list = cgm_from_galaxy_igmsystems(galaxy, igmsys, chk_lowz=False)
     assert len(cgm_list) == 1
     np.testing.assert_allclose(cgm_list[0].rho.value, 127.8324005876)
+
+def test_cgm_from_igmsystems_lowz():
+    # Load sightlines
+    sl_file = pyigm.__path__[0]+'/data/sightlines/Blind_CIV/J115120.46+543733.08.json'
+    igmsl = IGMSightline.from_json(sl_file)
+    igmsys = igmsl.make_igmsystems(vsys=400*u.km/u.s)
+    # Galaxy
+    galaxy = Galaxy((178.84787, 54.65734), z=0.00283)
+    # Go
+    cgm_list = cgm_from_galaxy_igmsystems(galaxy, igmsys, chk_lowz=True)
+    assert len(cgm_list) == 1
+    np.testing.assert_allclose(cgm_list[0].rho.value, 189.01377)
--- a/pyigm/cgm/tests/test_init_cgm.py
+++ b/pyigm/cgm/tests/test_init_cgm.py
@@ -72,7 +72,7 @@ def test_cgm_from_igmsystems():
     np.testing.assert_allclose(cgm_list[0].rho.value, 127.8324005876)
 
 def test_cgm_from_igmsystems_lowz():
-    # Load sightlines
+    # Load sightlines 
     sl_file = pyigm.__path__[0]+'/data/sightlines/Blind_CIV/J115120.46+543733.08.json'
     igmsl = IGMSightline.from_json(sl_file)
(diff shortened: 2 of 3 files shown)


pyigm/cgm/cgm.py
```diff
@@ -137,8 +137,8 @@ def __init__(self, galaxy, igm_sys, cosmo=None, name=None, rho=None, PA=None,
         self.igm_sys = igm_sys
 
         # Raise error for redshifts not in the Hubble flow
-        if (galaxy.z < 0.05) and chk_lowz:
-            raise NotImplementedError("Not prepared for such low redshift.  Need to implement corrections.")
+        #if (galaxy.z < 0.05) and chk_lowz:
```
profxj: We can likely remove these lines

pyigm/cgm/tests/test_init_cgm.py
```diff
@@ -70,3 +70,15 @@ def test_cgm_from_igmsystems():
...
+
+def test_cgm_from_igmsystems_lowz():
+    # Load sightlines 
+    sl_file = pyigm.__path__[0]+'/data/sightlines/Blind_CIV/J115120.46+543733.08.json'
+    igmsl = IGMSightline.from_json(sl_file)
+    igmsys = igmsl.make_igmsystems(vsys=400*u.km/u.s)
+    # Galaxy
+    galaxy = Galaxy((178.84787, 54.65734), z=0.00283)
+    # Go
+    cgm_list = cgm_from_galaxy_igmsystems(galaxy, igmsys, chk_lowz=True)
```
profxj: I think we can eliminate the chk_lowz here and throughout the code.

pyigm/cgm/utils.py
```diff
@@ -42,8 +42,14 @@ def calc_rho(galaxy, igm_sys, cosmo, ang_sep=None):
 
     if ang_sep is None:
         ang_sep = igm_sys.coord.separation(galaxy.coord).to('arcsec')
-    kpc_amin = cosmo.kpc_comoving_per_arcmin(galaxy.z)  # kpc per arcmin
-    rho = ang_sep.to('arcmin') * kpc_amin / (1+galaxy.z)
+    # Handle cases where object's distance needs correction from peculiar velocities
+    if (galaxy.z < 0.05) and chk_lowz:
```
profxj: Again, remove the chk_lowz.  

Please add a doc above stating how things are different when z<0.05

pyigm/cgm/utils.py
```diff
@@ -42,8 +42,14 @@ def calc_rho(galaxy, igm_sys, cosmo, ang_sep=None):
...
+        velcorrdict = velcorr_mould(galaxy,cosmo=cosmo)
+        kpc_amin = velcorrdict['scale'].to(u.kpc/u.arcmin)
```
profxj: I take it this one is a proper distance instead of comoving (hardly matters at such low z)

pyigm/cgm/utils.py
```diff
@@ -95,5 +101,122 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, R_max=300*u.kpc, dv_max=400*u
...
+
+    # Needed parameters
+    q0 = omega0 / 2.0 - omega_lambda
+    gamma = 2.0
+
+    # Load info from Mould+ 2000 Table 1A
+    clnames = ['Virgo', 'GA', 'Shapley']
+    clra1950 =  ['12h28m19s','13h20m00s','13h30m00s']
+    cldec1950 = ['+12:40:00','-44:00:00','-31:00:00']
+    clcoords = SkyCoord(clra1950,cldec1950,unit=(u.hourangle,u.degree),frame='fk4',equinox='B1950.0')
```
profxj, line 153: B1950 -- Nice!

pyigm/cgm/utils.py
```diff
@@ -95,5 +101,122 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, R_max=300*u.kpc, dv_max=400*u
...
+    clrangehi = np.array([2300., 6600., 16000.]) * u.km/u.s
+
+    # Convert B1950 coordinates (as given) and
+    clcoords = clcoords.icrs
+
+   # Convert input coords to Galactic coords
+    galcoords_gal = galaxy.coord.transform_to(frame='galactic')
+
+    # Transform to local group frame
+    c = 2.99792458e5 * u.km/u.s
```
profxj: Could use astropy.constants.c

pyigm/cgm/utils.py
```diff
@@ -95,5 +101,122 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, R_max=300*u.kpc, dv_max=400*u
...
+    v_LG = c * galaxy.z - 79.0 * (u.km/u.s) * np.cos(galcoords_gal.l).value \
+        * np.cos(galcoords_gal.b).value + 296.0 * (u.km/u.s) * np.sin(galcoords_gal.l).value \
```
profxj: Seems odd to see the .value attributes on the sin and cos, 
but I presume this is done right.

pyigm/cgm/utils.py
```diff
@@ -95,5 +101,122 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, R_max=300*u.kpc, dv_max=400*u
...
+    r0a = np.sqrt(v_LG ** 2 + clLGvel ** 2 - 2. * v_LG * clLGvel * costheta)
+
+    # Determine if object is in direction of one of the attractors.  If not, calculate velocity!
+    virgo = 0
+    GA = 0
+    shapley = 0
+
+    if (theta[0] < clrad[0]) & (
+        ((clLGvel[0] - r0a[0]) > clrangelo[0]) | ((clLGvel[0] + r0a[0]) < clrangehi[0])):
+        virgo = 1
```
profxj: I might use True instead of 1 but this is fine.

pyigm/cgm/utils.py
```diff
@@ -95,5 +101,122 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, R_max=300*u.kpc, dv_max=400*u
...
+    if (theta[1] < clrad[1]) & (
+        (clLGvel[1] - r0a[1] > clrangelo[1]) | (clLGvel[1] + r0a[1] < clrangehi[1])):
+        GA = 1
+    if (theta[2] < clrad[2]) & (
+        (clLGvel[2] - r0a[2] > clrangelo[2]) | (clLGvel[2] + r0a[2] < clrangehi[2])):
+        shapley = 1
+
+    if virgo or GA or shapley:
+        v_infall = np.zeros(3) * u.km/u.s
+        if virgo:
```
profxj: can more than 1 of this hit?

If not, you might use elif.

pyigm/cgm/utils.py
```diff
@@ -95,5 +101,122 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, R_max=300*u.kpc, dv_max=400*u
...
+        v_infall = fidvel * (costheta + (v_LG - clLGvel * costheta) / r0a * (r0a / clLGvel) ** (1. - gamma))
+        v_cosmic = v_LG + np.sum(v_infall)
+        flag = 0
+
+    # Derive remaining parameters to report
+    z_cosmic = v_cosmic / c
+    propdist = c / H0 * (z_cosmic - z_cosmic ** 2 / 2. * (1. + q0))
+    lumdist = propdist * (1. + z_cosmic)
+    angdiamdist = lumdist / (1. + z_cosmic) ** 2
+    distmod = 5. * np.log10(lumdist.to(u.pc).value) - 5.
```
profxj, line 214: Should this be 10pc?  Just checking..

profxj (changes requested):
Just a few changes requested.
Thanks a ton for doing this.

pyigm/cgm/cgm.py
jnburchett: yep

pyigm/cgm/tests/test_init_cgm.py
jnburchett: I'd rather not remove the chk_lowz altogether, as some people are made a little uneasy by this correction (e.g., the referee of paper 3).  Also, there are other formalisms for applying this correction, although the one we adopt is the most solid I'm aware of.

pyigm/cgm/utils.py
jnburchett: Added inline comment

pyigm/cgm/utils.py
jnburchett: Haha, it's what Mould et al. provide in their paper, so I just kept it consistent.

pyigm/cgm/utils.py
jnburchett: As you wish

pyigm/cgm/utils.py
jnburchett: Perhaps this is a bit awkward style.  I've moved the .value inside the function argument.

pyigm/cgm/tests/test_init_cgm.py
profxj: ok, but we should probably use a different flag, e.g.
  correct_lowz
or something like that.

pyigm/cgm/utils.py
jnburchett: Changed

pyigm/cgm/utils.py
jnburchett: I think, in general, multiple hits could occur, and the next little code block imposes that Shapley supercedes GA supercedes Virgo.

pyigm/cgm/utils.py
jnburchett: With the distance normalization inside the logarithm, it looks like:  
DM = 5 log_10(d / 10pc)
Should be right.

pyigm/cgm/tests/test_init_cgm.py
jnburchett: Cool.  I've refactored to change the argument/variable name and updated the CGM class definition docstring.

pyigm/cgm/utils.py
jnburchett: Turns out that moving the .value inside the parentheses would feed np.cos degrees when it take radians.  Moved .value back outside.

pyigm/cgm/utils.py
profxj: Got it.

profxj (approved):

profxj:
Merging.

PR #89, opened 2017-02-14
Fitdla GUI

profxj:
Adds a DLA fitting GUI to pyigm

Replicates much of what was in XIDL in x_fitlda and is somewhat improved.

Includes a new script:  pyigm_fitdla

Docs but no tests

### DIFF
--- a/docs/continuum.rst
+++ b/docs/continuum.rst
@@ -1,8 +1,8 @@
 .. highlight:: rest
 
 *********
-Continuum
-*********
+Continua
+********
 
 .. index:: contniuum
 
--- a/docs/fitdla.rst
+++ b/docs/fitdla.rst
@@ -1,0 +1,104 @@
+.. highlight:: rest
+
+.. _fitdla:
+
+***********
+XFitDLA GUI
+***********
+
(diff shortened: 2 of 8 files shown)


mrafelski (commented):
Code looks fine to me. Did some more testing, and a few more thoughts: 

1) If you hit S, it smooths, but breaks other things. If you hit s, it also has some functionality. Anyway, doing so breaks things. 

2) It would be nice to be able to specify an error array file. 

3)  I really do miss the functionality of saving a plot and of adding uncertainties. 

4) I get warning messages not too infrequently, do you as well? Both at start up, and at various times when using. 

5) The Q option doesn't really do anything, just a message to the screen.

profxj:
In response to Marc's next 5:

1.  Smoothing is too painful to implement.  Have turned it off to prevent the crashing.
2.  lt_readspec should be grabbing the error file automagically (it is for me).  Name it properly and it will work for you.
3.  Save with "P";  Uncertainties are on now and modified with U/u
4.  Ignore warning messages.  That is what they are there for
5.  Q now works

PR #93, opened 2017-03-12
Fit LLS GUI

profxj:
Ported the GUI from xastropy and refactored to PyQt5

Tested a bit but not enough.

Docs included


profxj:
This is ready for review now @astronomeara and @mifumagalli

astronomeara:
Can verify 'F' does not crash if z >> z_em

profxj:
That does not a code review qualify as

profxj:
Merging.

Issue #94, opened 2017-03-16
Statistical sample not correct in DLA surveys from literature


mneeleman:
When loading a DLASurvey from literature with the 'stat' sample e.g.,
`xq100=DLASurvey.load_XQ100(sample='stat')`
The output will show that _no_ DLAs were found:
`<IGMSurvey: nsys=0, type=DLA, ref=XQ-100, nsightlines=100>`
I think the issue is in DLASurvey line 440. 
`if np.abs(idla.zem-qsos['ZEM'][idx[qq]]) < zem_tol:`

Here the code is trying to compare the QSO emission from the QSOs table (i.e., the file pyigm_path+'/data/DLA/XQ-100/XQ100_zpath.fit') to the DLA table (i.e., pyigm_path+'/data/DLA/XQ-100/XQ100_abs.fit'). The latter however does not contain any information on the emission of the QSOs, so idla.zem=0 for all DLAs, and the conditional in the if statement is always False.

As a note; this redshift comparison is rather unnecessary anyway (at least for the literature sample). Indeed, it probably is causing more harm than good. If the wrong QSO was found by line 434:
`idx, d2d, d3d = match_coordinates_sky(dla_coord, qsos_coord, nthneighbor=1)`
then this does not mean at all that the DLA is not statistical. It just means that the wrong QSO was found.

profxj:
Is it just XQ-100?  And what if you don't specify stat?

profxj:
Addressed by PR #96 .

PR #96, opened 2017-04-11
Fix DLA survey stats

profxj:
Addresses bug identified by @mneeleman in Issue #94 

Updated XQ-100 notebook too.

### DIFF
--- a/docs/datasets/XQ100_systems.ipynb
+++ b/docs/datasets/XQ100_systems.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


mneeleman:
OK, that seems to work. However, the problem is indeed for all literature samples (G09 and P03 besides XQ100). Judicious placement of "skip_zem=True" should take care of that.

mneeleman (commented):
OK, that seems to work. However, the problem is indeed for all literature samples (G09 and P03 besides XQ100). Judicious placement of "skip_zem=True" should take care of that.

profxj:
Extended.  WIll merge if most tests pass.

profxj:
Failing some tests, but I'm on it.
Merging

PR #97, opened 2017-04-15
Refactor for linetools qtable -> table

profxj:
As titled.

All tests are assing on my laptop with Python 3.6
but Travis is having 'memory' issues.  No idea why..

PR #100, opened 2017-04-18
IGMGuesses and Qt5

profxj:
Quite a few edits to refactor for PyQt5.

Also had to handle the linetools refactor away from QTables.

igmguesses is working fine for me on Ubuntu and Python 3.6


profxj:
@ntejos should review *after* PR #97 

@jessicawerk should be able to run smoothly again provided 
she changes over to PyQt5.

pyigm/guis/igmguesses.py
```diff
@@ -1421,13 +1435,13 @@ def __init__(self, component=None, parent=None):
         self.parent = parent
         #if not status is None:
         #    self.statusBar = status
-        self.label = QtGui.QLabel('Component:',self)
+        self.label = QLabel('Component:',self)
         self.zwidget = ltgsm.EditBox(-1., 'zc=', '{:0.5f}')
```
ntejos, line 1453: There is an "unwanted" behaviour with ``EditBox`` now. Basically, when I edit, say the b-value to ``10``, I need to write "1" and "0". Right after I write "1" in the box, the fit is updated to ``b=1`` km/s before letting me put the "0". This is somewhat annoying.

pyigm/guis/igmguesses.py
```diff
@@ -1442,37 +1456,33 @@ def __init__(self, component=None, parent=None):
...
-        self.connect(self.zwidget.box, 
-            QtCore.SIGNAL('editingFinished ()'), self.setbzN)
-        self.connect(self.bwidget.box, 
-            QtCore.SIGNAL('editingFinished ()'), self.setbzN)
-        self.connect(self.Cwidget.box, 
-            QtCore.SIGNAL('editingFinished ()'), self.setbzN)
+        self.Nwidget.box.textChanged[str].connect(self.setbzN)
+        self.zwidget.box.textChanged[str].connect(self.setbzN)
+        self.bwidget.box.textChanged[str].connect(self.setbzN)
+        self.Cwidget.box.textChanged[str].connect(self.setbzN)
```
ntejos: May be related to this(these) new call(s).

pyigm/guis/igmguesses.py
```diff
@@ -1510,11 +1522,13 @@ def reset(self):
...
     def update_component(self):
-        '''Values have changed'''
+        """Values have changed"""
+        self.update = False  # Avoids a bit of an internal loop
         self.Nwidget.set_text(self.component.attrib['logN'])
         self.zwidget.set_text(self.component.attrib['z'])
         self.bwidget.set_text(self.component.attrib['b'].value)
         self.Cwidget.set_text(self.component.comment)
+        self.update = True  # Avoids a bit of an internal loop
         if self.parent is not None:
```
ntejos, line 1556: Or may be related to this new way to ``update_component()``. Seems we need a way to tell the GUI that the user is done putting strings into the boxes before setting ``self.update = True``

pyigm/guis/igmguesses.py
ntejos: Also, if I remove all the values, i.e. leaving only the string ``""``, the GUI breaks because cannot convert that string to a float. The same happens if I accidentally add a string that would make ``float(str)`` fail.
Update: I added a commit that deals with this, but the first issue remains (see previous comment).

pyigm/guis/igmguesses.py
```diff
@@ -1525,16 +1544,22 @@ def set_label(self):
...
-            self.component.attrib['b'] = (float(self.bwidget.box.text()))*u.km/u.s
-            self.component.comment = str(self.Cwidget.box.text())
+            try:
+                self.component.attrib['logN'] = (float(self.Nwidget.box.text()))
+                self.component.attrib['z'] = (float(self.zwidget.box.text()))
+                self.component.attrib['b'] = (float(self.bwidget.box.text()))*u.km/u.s
+                self.component.comment = str(self.Cwidget.box.text())
+            except ValueError:  # this is when the str cannot be converted to float
+                print("The new value is not valid. Try again.")
             #QtCore.pyqtRemoveInputHook()
```
ntejos: This bit here was added by me to address a previous issue; please review.

pyigm/guis/igmguesses.py
```diff
@@ -1634,31 +1659,62 @@ def remove_item(self,comp_name):
...
+        comp = AbsComponent.from_abslines(abslines, stars=stars)
+    else:  # H2
+        Zion = (-1, -1)  # code for molecules
+        Ntuple = (17, -1, 1)  # initial guess for Ntuple (needs to be given for adding lines from linelist; see below)
+        comp = AbsComponent(zero_coord, Zion, z, vlim, Ntup=Ntuple)
+        # add handy attrib
+        comp.attrib['init_wrest'] = wrest
+        # add abslines
+        comp.add_abslines_from_linelist(llist='H2', wvlim=(wvmin, wvmax), min_Wr=None)
+        # QtCore.pyqtRemoveInputHook()
```
ntejos, line 1736: I also added this bit from a while ago (I had forgotten about this commit), but seems harmless here (only applies for LineList H2).

ntejos (approved):
This seems to be working, although some notes before merging:
- Check my comment about the annoying behaviour in update_component(), in case you see an easy fix.
- Check the couple of commits I added to solve a simple issue + an old one that deals with H2 LineList (unrelated to this PR, but harmless)

pyigm/guis/igmguesses.py
profxj: Yeah, this is the new functionality of EditBox.  If you can figure
out how to stop it, we should change it.

pyigm/guis/igmguesses.py
profxj: These new calls are necessary in PyQT5.  I think

pyigm/guis/igmguesses.py
profxj: Trust me, I fussed with this quite a bit.

Things were crashing a lot previously..

pyigm/guis/igmguesses.py
profxj: ok

pyigm/guis/igmguesses.py
ntejos: It seems to me that we would need to create a button on the widget like "Update", for when the user is done editing the box(es) the values are retrieved and updated after pressing such button. What do you think?

profxj: Feel free to do so.  
Good luck..

ntejos:
Here is an attempt at the "setzbNc" issue mentioned before. I have now added a button "Update" that only when pressed the values are updated from the values given by the user in the widget(s). This now includes the "comment" widget (because the annoying behaviour was also present in comment). 

TODO: Redefine the layout for the "update button", ideally has to go below all four "z", "N", "b", "comment" to make it implicit that the comment needs to be updated using the button too. When I tried to move the button around, I lost connectivity to it and the button was not longer working.

PR #101, opened 2017-04-21
Pickle and a bit more

profxj:
Several modifications to handle Python 3.


mifumagalli (approved):
Looks fine - unsure on the need for L83-85, but ok otherwise.

PR #103, opened 2017-04-25
CGM Modeling

profxj:
New object for modeling the CGM.

Based on the formalism described in Mathews & Prochaska (2017).

Includes tests and basic Notebook.


pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+        # Return
+        return ne
+
+    def nH(self, xyz):
+        """ Calculate the Hydrogen number density
+        Includes a correction for Helium
+
+        Parameters
+        ----------
+        xyz
```
jnburchett: Definition?

docs/cgm.rst
```diff
@@ -151,3 +152,11 @@ are included in the data directory.  You can read in the dataset with::
...
 
 
+CGMModel
+========
+
+Simple models designed to reproduce observations of the CGM
+may be generated using the CGMModel class.  Current implementation
+is based on the Mathews & Prochaska (2017) halo profile.
+
+See the CGM_Models Notebook for examples.
```
jnburchett, line 162: For some reason, I don't see this notebook.  Does the final commit needs to be pushed?

pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+from astropy import units as u
+from astropy import constants as const
+
+try:
+    basestring
+except NameError:  # For Python 3
+    basestring = str
+
+def rad3d2(xyz):
+    return xyz[0]**2 + xyz[1]**2 + xyz[-1]**2
```
jnburchett: Is there a reason that the final term isn't 'xyz[2]**2'?

pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def setup_param(self):
+        """ Setup key parameters of the model
+        """
+        # Cosmology
+        self.H0 = 70. *u.km/u.s/ u.Mpc
+        self.fb = 0.16       # Baryon fraction
+        self.rhoc = 9.2e-30 * u.g / u.cm**3
+        # Dark Matter
+        self.r200 = (((3*10**self.M_halo * const.M_sun.cgs) / (4*np.pi*200*self.rhoc))**(1/3)).to('kpc')
```
jnburchett: Need to make whole numbers floats

pyigm/cgm/models.py
jnburchett: Also, I notice that when this is called, you immediately go ahead to take the sqrt().  Are there cases where you *wouldn't* want the square root?  If not, go ahead and put the sqrt() in the function?

jnburchett (changes requested):
The CGM_Models notebook seems to be missing.  Once I can see that, I'll do another in-depth review.

pyigm/cgm/models.py
profxj: Hmm.  Must be on my other laptop. Will 
upload it when I return from Tokyo

docs/cgm.rst
```diff
@@ -151,3 +152,11 @@ are included in the data directory.  You can read in the dataset with::
 See the Notebook in examples for further details.
 
 
+CGMModel
+========
+
+Simple models designed to reproduce observations of the CGM
+may be generated using the CGMModel class.  Current implementation
+is based on the Mathews & Prochaska (2017) halo profile.
```
yzhenggit, line 160: This Mathews & Prochaska (2017) paper is not out yet, right? Do you need a second person to double check the equations and the astropy unit calculation?

pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+        # Cut at rmax*rvir
+        if Rperp > rmax*self.r200:
+            return 0. / u.cm**2
+        # Generate a sightline to rvir
+        zmax = np.sqrt(self.r200 ** 2 - Rperp ** 2).to('kpc')
+        zval = np.arange(-zmax.value, zmax.value+dz, dz)  # kpc
+        # Set xyz
+        xyz = np.zeros((3,zval.size))
+        xyz[0, :] = Rperp.to('kpc').value
+        xyz[2, :] = zval
```
yzhenggit, line 247: @profxj Here you assume the origin is at (x, y, z)=(0, 0, 0), right? OK if it's only for 1D NFW profile modeling. But would be good to add in an origin values of (x0, y0, z0)=(0, 0, 0) for future functions. Not sure how useful that would be, but I'm thinking about some cosmological simulation boxes where people usually need to define a center before shooting sightline array.

pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # Cut at rmax*rvir
+        if Rperp > rmax*self.r200:
+            return 0. / u.cm**2
+        # Generate a sightline to rvir
+        zmax = np.sqrt(self.r200 ** 2 - Rperp ** 2).to('kpc')
+        zval = np.arange(-zmax.value, zmax.value+dz, dz)  # kpc
+        # Set xyz
+        xyz = np.zeros((3,zval.size))
+        xyz[0, :] = Rperp.to('kpc').value
```
yzhenggit, line 246: @profxj (this is purely a suggestion): Here xyz[1, :]=0 since it's based on 1D profile along r; for future analyses, maybe it would be helpful to define a position angle _theta_ now, which could be used to determine the x/y axis value of the sightline array on a polar coordinate system. In such case, xyz[0, :]=Rperp\times np.cos(theta)+x0, xyz[1, :]=Rperp\times np.sin(theta)+y0, where (x0, y0) is the origin of the coordinate system (see my next comment). In this case, we will be able to probe the different location of the halo if the halo profiles are 2D or even 3D. For now, the position angle can be set as 0 in the code, so the result would be exactly the same.  :)

pyigm/cgm/models.py
yzhenggit: Radius calculation would be misleading if xyz origin is not defined as (0, 0, 0).

pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+        txt += '   Phases modeled = {} \n'.format(self._pdict.keys())
+        txt += '>'
+        return (txt)
+
+class CGMPhase(object):
+    """ Model of a single phase of the CGM
+    Parameters:
+    -----------
+    phase : str
+      'hot'
```
yzhenggit: What's your temperature definition of "hot"?

pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+        # Init more
+        self.setup_param()
+
+    def setup_param(self):
+        """ Setup key parameters of the model
+        """
+        # Cosmology
+        self.H0 = 70. *u.km/u.s/ u.Mpc
+        self.fb = 0.16       # Baryon fraction
+        self.rhoc = 9.2e-30 * u.g / u.cm**3
```
yzhenggit, line 139: I believe references for these numbers are in Mathews & Prochaska (2017), but maybe some short comments here listing where they come from so that people could easily figure out the related papers in the future?

pyigm/cgm/models.py
```diff
@@ -0,0 +1,240 @@
...
+    alpha : float, optional
+      Parameter to modify NFW profile power-law
+    y0 : float, optional
+      Parameter to modify NFW profile position
+    """
+    def __init__(self, M_halo=12.2, c=7.67, f_hot=0.9, alpha=0., y0=1., **kwargs):
+        # Init
+        CGMPhase.__init__(self, 'hot')
+        # Param
+        self.M_halo = M_halo
```
yzhenggit: Is it because M_halo is in log so you don't assign the unit const.M_sun.cgs here but do it in line 129? Have a feeling that astropy unit maybe the code killer at some point ...

profxj:
Just pushed the Notebook.  Will be addressing the other comments shortly
Thanks!

pyigm/cgm/models.py
```diff
@@ -0,0 +1,241 @@
...
+
+        Returns
+        -------
+        nH : float or ndarray
+          Hydrogen number density in cm**-3
+        """
+        pass
+
+
+class ModifiedNFW(CGMPhase):
```
jnburchett, line 103: If this is really a 'straight NFW' as indicated in the notebook, why is it called 'ModifiedNFW'?

pyigm/cgm/models.py
jnburchett: Still should make these floats

pyigm/cgm/models.py
jnburchett: Need to add `ndarray
          Position assumed in kpc`

pyigm/cgm/models.py
```diff
@@ -0,0 +1,241 @@
...
+        xyz = np.zeros((3,zval.size))
+        xyz[0, :] = Rperp.to('kpc').value
+        xyz[2, :] = zval
+
+        # Integrate
+        ne = self.ne(xyz) # cm**-3
+        Ne = np.sum(ne) * dz * 1000  # pc cm**-3
+
+        # Return
+        return Ne * u.pc / u.cm**3
```
jnburchett, line 254: What's up with these units?  Perhaps default to cm^-2 because most CGM people think in these terms?

jnburchett (changes requested):

pyigm/cgm/models.py
yzhenggit: Oh, pc cm-3 is the unit for dispersion measure. It tells the electron number density (cm-3) along the distance in pc.

jnburchett:
I get that but I just wonder if we should default to those units.  

-Joe

pyigm/cgm/models.py
profxj: Notebook now uploaded

pyigm/cgm/models.py
profxj: Python uses float math once any float is included.

profxj: But I will add one on the 10

pyigm/cgm/models.py
profxj: Added doc on xyz origin.

pyigm/cgm/models.py
profxj: Updated

docs/cgm.rst
profxj: Pushed

docs/cgm.rst
profxj: Good suggestion, but I think we are ok on this.

pyigm/cgm/models.py
profxj: Good suggestion, but I think I'll leave that for a future PR (i.e. when I need it).

pyigm/cgm/models.py
profxj: Also an excellent suggestion.  The ne2001 code that I link to
in my own analysis works on xyz, so I prefer to leave that.  But
I will eventually add a position angle approach as well.

pyigm/cgm/models.py
profxj: Added a doc stating > 10^6 K.  Of course, this is rather arbitrary.

pyigm/cgm/models.py
profxj: These are all standard cosmological values.

pyigm/cgm/models.py
profxj: Good point.  Am redifining this as log_Mhalo
and will carry around a proper Mhalo too.

pyigm/cgm/models.py
profxj: Well, I suppose most properly it is a 'modifiableNFW' but 
I think I'm going to keep it as ModifiedNFW.

pyigm/cgm/models.py
profxj: Right.  This is the 'standard' unit

profxj:
Thanks for all comments @jnburchett and @yzhenggit 

Have another look please.  Thanks!

profxj:
merging.

PR #104, opened 2017-04-28
Igmguesses reliability

ntejos:
Reliability attribute now consistent with the nes linetools' definition of AbsComponent


profxj:
fine by me.

Merge if this has already cleared in linetools.

PR #110, opened 2017-05-20
H100 metallicities

profxj:
Ingests metallicity values and kinematics from 
Table 2 of Neeleman+13.


pyigm/surveys/dlasurvey.py
```diff
@@ -164,33 +168,41 @@ def load_H100(cls, grab_spectra=False, load_sys=True, isys_path=None):  #skip_tr
...
+                pdb.set_trace()
+            # Metallicity
+            dla_survey._abs_sys[mt[0]].ZH = tbl2['[M/H]'][ii]
+            dla_survey._abs_sys[mt[0]].sig_ZH = tbl2['e_[M/H]'][ii]
+            if tbl2['M'][ii] in ['S','Si','O']:
+                dla_survey._abs_sys[mt[0]].flag_ZH = 1  # Alpha
+            elif tbl2['M'][ii] in ['Zn']:
+                dla_survey._abs_sys[mt[0]].flag_ZH = 2  # Zn
+            elif tbl2['M'][ii] in ['Fe']:
+                dla_survey._abs_sys[mt[0]].flag_ZH = 4  # Zn
```
mneeleman: Probably want to change the comment to # Fe

pyigm/surveys/dlasurvey.py
```diff
@@ -164,33 +168,41 @@ def load_H100(cls, grab_spectra=False, load_sys=True, isys_path=None):  #skip_tr
...
+        for ii, iqso, izabs in zip(range(len(tbl2)), tbl2['QSO'], tbl2['zabs']):
+            mt = np.where((qsonames == iqso) & (np.abs(izabs-zabs) < 1e-3))[0]
+            if len(mt) == 0:
+                pdb.set_trace()
+            elif len(mt) != 1:
+                pdb.set_trace()
+            # Metallicity
+            dla_survey._abs_sys[mt[0]].ZH = tbl2['[M/H]'][ii]
+            dla_survey._abs_sys[mt[0]].sig_ZH = tbl2['e_[M/H]'][ii]
+            if tbl2['M'][ii] in ['S','Si','O']:
```
mneeleman, line 192: Not sure why you are giving this a number value. Why not just report the metal used for the determination? Still many people have issue with Si as a metallicity determination, now all alpha elements get a flag of 1 and it becomes impossible to separate Si from S and O. If you insist on the number system, I would give each element a different number.

profxj: Am trying to make it somewhat backwards compatible to
what we had in the old days..  

But good point.  Am now also saving the actual element too.

pyigm/surveys/dlasurvey.py
profxj: done

profxj:
merging.

Issue #111, opened 2017-05-22
Include more meta data in IGMGuesses output file


profxj:
Include:

sl_dict['RA'] = [0.]*u.deg     # J2000, decimal degrees
sl_dict['DEC'] = [0.]*u.deg    # J2000, decimal degrees
sl_dict['zem'] = [0.]   # Emission redshift
sl_dict['Instrument'] = ['e.g. HST/COS/G130M PADDING']   # Telescope/Instrument/Grating
sl_dict['Creator'] = ['SOMEONES_NAME']   # Emission redshift

Facilitates Database generation

ntejos:
Addressed in PR#112

PR #112, opened 2017-05-23
IGMGuesses: adding and storing metadata

ntejos:
As stated. A new dictionary of metadata is kept in `self.meta` attribute an written into the .json file with the keyword `"meta"`, having the following info as example:
```
"meta": {
        "DEC_deg": 66.0,
        "RA_deg": 1234.0,
        "creator": "John Smith",
        "instrument": "HST/COS/G130M",
        "object": "PG0953",
        "zem": 0.237
```
These values are initially filled with placeholders, in which case IGMGuesses will ask the user to provide the correct ones via prompt. 

This PR addresses Issue #111

### DIFF
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -1988,7 +1988,7 @@ def fill_meta(meta):
             radec = radec.split(",")
             meta["RA"] = radec[0]
             meta["DEC"] = radec[1]
-    # JNAME
+    # JNAME (will also check whether coordinates are sensible)
     coord = SkyCoord(meta['RA'], meta['DEC'], unit='deg')
     jname = ltu.name_from_coord(coord, precision=(2, 1))
     meta['JNAME'] = jname
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -1972,7 +1972,7 @@ def fill_meta(meta):
     Returns
     -------
     meta_updated : dict
-        A new meta dictionary with values that are not default
+        A new meta dictionary with values given by user
 
     """
     meta_ref = init_meta()
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -10,7 +10,7 @@
 #;   14-Aug-2015 by JXP
 #;-
 #;- NT: New version using linetools' AbsComponent
-#;-
+#;- NT: New version with enhancements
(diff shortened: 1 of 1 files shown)


pyigm/guis/igmguesses.py
```diff
@@ -1929,7 +1943,68 @@ def from_igmguesses_to_joebvp(infile, outfile):
...
+    -------
+    meta : dict
+
+    """
+    meta = dict()
+    meta['RA_deg'] = 0.
+    meta['DEC_deg'] = 0.
+    meta['object'] = "unknown"
+    meta['zem'] = 0.
+    meta['instrument'] = 'unknown'
```
profxj: Please capitalize these (like in z1IGM)

pyigm/guis/igmguesses.py
```diff
@@ -1929,7 +1943,68 @@ def from_igmguesses_to_joebvp(infile, outfile):
...
+def init_meta():
+    """Creates a general meta dictionary
 
+    Returns
+    -------
+    meta : dict
+
+    """
+    meta = dict()
+    meta['RA_deg'] = 0.
```
profxj: Please just use RA and DEC

pyigm/guis/igmguesses.py
profxj: Add Date either here or elsewhere

profxj (changes requested):

pyigm/guis/igmguesses.py
ntejos: How about we add the date automatically? Based on the date the .json file is written?

ntejos:
@jnburchett @jessicawerk and others, please feel free to comment on this PR.

pyigm/guis/igmguesses.py
profxj: That would be great.  You can see code in the z1igm PR that
you can copy.

pyigm/guis/igmguesses.py
ntejos: Sure.

pyigm/guis/igmguesses.py
ntejos: Sure

ntejos: Roger.

ntejos:
Ready for review again @profxj

profxj:
Looks great!

Merging

PR #113, opened 2017-05-24
Igmguesses enhance

ntejos:
Asks the user for minimum equivalent width, and minimum velocity limit for AbsLines, when using the "hidden feature" `'@'` for writing .joebvp outputs.


profxj:
Assuming you have tested it yourself, 
go ahead and merge.

PR #114, opened 2017-05-24
deepcopy for writing out components.

ntejos:
Just that, but it will be important.

### DIFF
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -1886,7 +1886,7 @@ def blending_info(components, specfile, min_vlim=100*u.km/u.s, min_ew=0.005*u.AA
     print('IGMGuesses: computing blends between components, it may take a while...\n')
 
     # create a copy of component list that has a minimum vlim incorporated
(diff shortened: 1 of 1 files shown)


ntejos:
Merging.

pyigm/guis/igmguesses.py
```diff
@@ -1886,7 +1886,7 @@ def blending_info(components, specfile, min_vlim=100*u.km/u.s, min_ew=0.005*u.AA
     print('IGMGuesses: computing blends between components, it may take a while...\n')
 
     # create a copy of component list that has a minimum vlim incorporated
-    comps_copy = copy.copy(components)
+    comps_copy = copy.deepcopy(components)
```
profxj, line 1889: Is this an AbsComponent?

If so, we should generate a specific copy
method for AbsComponent.

ntejos: Its a list of AbsComponents. Indeed it would be nice to have the copy method in AbsComponent, but that would a linetools issue. I need this patch for our current workflow.

profxj:
Merging

PR #123, opened 2017-07-19
from_json method for CGMAbssys

profxj:
As titled.

About to add a test


profxj:
Am going to merge into @marijana777 's PR

PR #124, opened 2017-07-25
Fitlls with multi spec

profxj:
Refactored to handle more than 1 spectrum at a time.

A few other bits and pieces added too.

Requires PR #403 pending in linetools:

https://github.com/linetools/linetools/pull/403


profxj:
Merging.

PR #125, opened 2017-07-28
small fixes to handle with python3 compatibility

mifumagalli:
I am testing the code with new grids that include turbulence. In principle, this should work out of the box. However,  in the process, I made small adjustments (I think required for python 3.6). 

Note that the current version of the code fails to create a hdf5, which I believe is associated to this issue: https://stackoverflow.com/questions/23220513/storing-a-list-of-strings-to-a-hdf5-dataset-from-python

I did not find a permanent fix, so I am currently skipping hdf5 file i/o if fails.

### DIFF
--- a/pyigm/metallicity/mcmc.py
+++ b/pyigm/metallicity/mcmc.py
@@ -295,13 +295,12 @@ class Emceebones(object):
         -------
 
         """
-
-
+        
         #load the model in a format that can be handled later on
-        fil=open(model)
+        fil=open(model,'br')
         modl=pickle.load(fil)
         fil.close()
-
+    
         #unpack axis tag, axis value, grid column, grid ions
         self.mod_axistag=modl[0]
         self.mod_axisval=[]
@@ -326,14 +325,12 @@ class Emceebones(object):
 
         #loop over all the ions observed
         for obs in data:
-
             # Check for zero error (causes unexepcted failure)
             if obs[2] <= 0.:
                 raise ValueError("Cannot have 0 error on the column density, even in a limit.  Fix {}".format(obs[0]))
 
             #check obs one by one for corresponding entries in model
-            if modl[2].has_key(obs[0]):
-
+            if (obs[0] in modl[2]):
                 #stack the observable
                 #[list of tuples for observations with (ion,column,error)]
                 self.data.append(obs)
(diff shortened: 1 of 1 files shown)


cwotta (approved):
(1) Forgot about 'b' for Windows machines, sorry (and dict.has_key()...).

(2) Given point (1), should the H5Py write mode also include 'b' (line 404)?

(3) Is this the issue you're facing with H5Py?  Looks like no fixes if we stick with H5Py -- we could potentially switch, but I don't know what that would entail.  https://stackoverflow.com/questions/32654665/python3-crash-when-creating-hdf5-file-via-h5py

profxj:
I really prefer hdf5 if we can keep it.  I think the issue here is 
likely unicode vs. str.   I ran into this issue in specdb and wrote
a simple method to deal with my issue.  See hdf_decode() in this
module:

https://github.com/specdb/specdb/blob/master/specdb/utils.py

If that fixes your trouble, let's stick with hdf5.  Thanks

profxj:
Sorry, my last comment was actually the other way, i.e. 
dealing with reading hdf5 which come in as bytes.

Below is a snippet of code I used to deal with writing to hdf5:

```
    tbl_keys = tbl.keys()
    for key in tbl_keys:
        # Check for unicode
        if 'U' in tbl[key].dtype.__repr__():
            warnings.warn("unicode in column {:s}.  Will convert to str for hdf5".format(key))
            tmp = Column(tbl[key].data.astype(bstr), name=key)
            tbl.remove_column(key)
            tbl[key] = tmp
```

where bstr is defined for Python 2/3 compatability as:

```
try:
    bstr = bytes
except NameError:  # For Python 2
    bstr = str
```

mifumagalli:
@cwotta: Changes are for python 3 on a mac (where I am currently developing). I have not encountered (so far) issues with binary flag in hdf5. Also, not sure b is a write option for hdf5. Skipping for now.

As for HDF5 issue, yes, encoding seems to be the issue. Note that I have not removed hdf5, but only added a safety check so that code does not crash but tries to continue if hdf5 fails. I'll let @cwotta develop a more permanent fix for that bit of the code.

profxj (approved):
Ok, I'm approving.

profxj:
merging.

PR #126, opened 2017-07-28
Handle cosmo differently in lyman_ew()

profxj:
Makes use of the cosmo in fn_model (if set)
instead of simply defaulting to a vanilla cosmo.

New test too.


betalusso (approved):

profxj:
Merging.

PR #128, opened 2017-09-02
typo fix and rename in "densToLogU"

cwotta:
Renamed function from "dens2logu" to "densToLogU".  Also fixed a typo in that function: it had called "logu2dens" (which does not exist, because it had previously been renamed "logUToDens").


profxj:
Looks fine. Merging

PR #130, opened 2017-09-03
Add metallicity to Notebook and more

profxj:
As titled


cwotta:
Thanks for updating with metallicity info.  Followed the steps, it worked for me.

PR #135, opened 2017-09-22
calc_rhoimpact: accept astropy Columns

jnburchett:
Getting a bit kludgy, but method will not work with PHL1377 galaxy table without it.


profxj:
Yeah, that is a bit messy.

And now that I think about it, I think there is a stand-alone
rho method.  Yup, in   pyigm.cgm.utils   calc_rho()

Ah, but that only does one galaxy/IGM pair at a time.  Still,
we might think about a way to avoid having both..

Your call @jnburchett 

Merge this if you prefer.

PR #136, opened 2017-09-25
Fitted DLA stats and more

profxj:
Enables fitting l(z) and several other quantities, mainly
in the context of DLAs.  Most of these are related to
an FRB analysis.

Adds a new set of DLA fitting parameterizations that
are loaded into DLASurvey by default.

Lots of docs updated.

Tests now passing Travis (except one which is unrelated to pyigm).


mneeleman (commented):
1) I don’t think DX should be listed as a column in the sightlines as it is dependent on the cosmology and therefore not a measurable quantity/property of the observations. That is if you change the cosmology that you want, this column does not seem to change:

from pyigm.surveys.dlasurvey import DLASurvey
from astropy.cosmology import WMAP9 as Wcosmo

sdss_stat = DLASurvey.load_SDSS_DR5()
sdss_stat.sightlines[0:5]
sdss_stat.cosmo = Wcosmo
sdss_stat.sightlines[0:5]

Both give the same DX.

2) For CII*, you assume that a flag of 6 means a detection from the two examples cases in the H100 sample (J0040-0915 and J1201+2117), it looks like it should be interpreted as junk (some forest line or many rest lines seem to be interfering with an accurate determination).

3) Maybe add in the documentation how to change the cosmology that is used.  Although I found it quickly, so it can’t be too hard.

4) It would probably be beneficial not just to be able to calculate rho_HI, but also to allow for calculation of Omega_HI. 

5) The uncertainty on the fitted values of fn, l(z) and ne/nH are calculated during the bootstrap procedure, but are these not stored?
they are not in .dla_fits
On a similar note, it would be good to return uncertainties on the fitted values as well.

6) line 566 in dlasurvey.py should be self.binned_lox(…

profxj:
Errors on lz now available for fit.

Updated nenH values from @mneeleman 

Passing tests on machine.

Ready to merge?

profxj:
Thanks @mneeleman --   Responses to yours are below:

1.  DX in that Table wasn't being used (I don't think).  But I have removed it nevertheless
2.  Good catch.  Have converted flag=6 to 0 now
3.  For which method(s)?
4.  Agreed, but I'm too lazy right now and deciding whether to correct for He always annoys me..
5.  Good one.  Am not sure how to propagate the uncertainies on the returned (fitted) values
but should be easy enough.   Let's discuss.  Am putting in the errors on the parameters now.  
For ne/nH, did you calculate uncertainties??
6.  Good catch

I forgot to hit the "Comment" button previously

mneeleman:
As to changing the cosmology, I meant the binned evaluations of DLASurvey. Something in the documentation that to change to a different cosmology one can set it using any cosmology as defined in the astropy.cosmology package. e.g.:

from astropy.cosmology import WMAP9 as WMAPcosmo
dla = DLASurvey(ref='null')
dla.cosmo = WMAPcosmo

On a related note, the fitted values should probably have a comment denoting what cosmology was used to calculate these fits. That is dla.dla_fits should list not only the reference paper, but also the cosmology.

PR #139, opened 2017-10-03
Refactor calc rho

profxj:
Refactors calculation of impact parameter to 
top level of pyigm.

Tests and other code were updated too.
No new docs.


profxj:
hold up.  want to swap ordering..

profxj:
Reopening.

pyigm/cgm/utils.py
```diff
@@ -15,73 +15,228 @@
...
     cosmo : astropy.cosmology
-    Galactic : bool, optional
-      Calculate for our Galaxy!
+    ang_sep : Quantity, optional
     correct_lowz : bool, optional
-      Apply corrections for the local universe, as desired
-      Follows Mould et al. 2000
-    ang_sep : Angle or Quantity
-      Input angular separation
-      May speed up calculation
```
jnburchett, line 36: Why remove all this?  Some potentially very useful information here...

jnburchett (approved):
Looks good except for couple docstring changes.

pyigm/cgm/utils.py
profxj: Decided to use **kwargs instead

profxj:
merging

PR #140, opened 2017-10-05
CGM utils for nondetections

jnburchett:
Adapt cgm.utils to allow for CGM systems to be paired even if no matching IGMSystem is already in the IGMSightline


profxj:
This PR conflicts greatly with #139 (I think).

Let's clear that @jnburchett then work on this one.  
Closing for now.

PR #141, opened 2017-10-05
CGM nondetections

jnburchett:
Enables instantiation of CGMAbsSys objects even when no measurements exist as IGMSystems; conflicts with PR #139 resolved


pyigm/cgm/utils.py
```diff
@@ -13,10 +13,14 @@
...
 
 from linetools import utils as ltu
+from linetools.lists.linelist import LineList
+from linetools.isgm.abscomponent import AbsComponent
+from linetools.spectralline import AbsLine
 
 from pyigm.field.galaxy import Galaxy
 from pyigm.abssys.igmsys import IGMSystem
 
+ismlist = LineList('ISM')
```
profxj: I'm not keen on defining this here.

Keep it in the method(s).

profxj (commented):
Looks good to me.

Just address the one comment and merge.

PR #144, opened 2017-10-12
Small changes to input file, and fix for Python2

cwotta:
M.Fumagalli changed from `fil=open(model,'r')` to `fil=open(model,'br')` for use with Python3.  But this breaks in Python2.  So now we do a try/except.

Allowed --wotta option to accept -grid="x" option.

Fixed example queue_script.sh, so command can follow symlinks.

### DIFF
--- a/pyigm/metallicity/mcmc.py
+++ b/pyigm/metallicity/mcmc.py
@@ -297,9 +297,16 @@ class Emceebones(object):
         """
         
         #load the model in a format that can be handled later on
-        fil=open(model,'br')
-        modl=pickle.load(fil)
-        fil.close()
+        try:
+            ##Python3
+            fil=open(model,'br')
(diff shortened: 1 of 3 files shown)


pyigm/metallicity/mcmc.py
```diff
@@ -297,9 +297,16 @@ def loadmodel(self, data, model):
         """
         
         #load the model in a format that can be handled later on
-        fil=open(model,'br')
-        modl=pickle.load(fil)
-        fil.close()
+        try:
```
profxj, line 300: Most of the lines in the try/except 
are repeated.   Please pull those out, i.e.

```
modl=pickle.load(fil)
fil.close()
```

profxj (commented):
small edits to make

PR #145, opened 2017-10-13
minor fixes

cwotta:
Fixed Python2 read-pickle-file error.
Allowed --wotta option to accept -grid="x" option.
Allowed example queue_script.sh to follow symlinks in "find".


cwotta:
@profxj : how do these changes look?  I've pulled out the repeated code

profxj:
Looks good.
Merging.

PR #147, opened 2017-10-16
Units fix for new astropy

profxj:
fixes igmguesses (I hope).


ntejos:
Looks good to me. We tested in Jess' computer and seemed fine. Happy to merge is everyones is ok with it.

AngryGroceries:
The fix works. It did not work with an older version of linetools that I had installed, but it works with the current linetools master and with the linetools voigt test 5622cc9 

edit: used 5622cc9 not a906010

profxj:
Merging.

PR #149, opened 2017-10-19
Load COS Dwarfs

profxj:
As titled.  One needs access to a private
file in Dropbox for the time being.


profxj:
@rongmon reports success.

merging

PR #150, opened 2017-10-20
add emission redshift in suptitle

ntejos:
Have emission redshift given in the main velocity window widget of igmguesses.


profxj:
Looks fine;  merging

PR #156, opened 2017-10-28
Refactor coords in igmsurvey

profxj:
Speeds things up remarkably.

First step on the way to a more substantial Refactor


jnburchett (approved):
I like this...would be nice to refactor cgmsurvey similarly!  

One observation, although I think it's unrelated:  I tried testing by loading HDLLS, but it takes forever!  It seems to continually reload the LineList, so I assume that's the issue but thought I would mention it just in case.

Merge if happy.

profxj:
merging

PR #160, opened 2017-11-30
COS-Halos v1.2

profxj:
Code to generate a v1.2 of the COS-Halos database.

Had forgotten about this branch, which includes the
Absolute r mag.

Once people are happy with it, I'll add in the 1.2 tarball
to the Repo.


profxj:
See the bottom of this Notebook:

https://github.com/pyigm/pyigm/blob/cos_abmagr/docs/datasets/COS-Halos_database.ipynb

for the new attribute.

marijana777:
It looks good to me, and it works. But, would it be possible to also add absolute u-magnitude, or u-r color?

profxj:
What do you think @jessicawerk ?

marijana777:
Actually, it might be useful if this could be fixed before merging, @profxj  :
From the new branch pyigm_igmguesses is giving an error : 
    from PyQt4 import QtGui
ModuleNotFoundError: No module named 'PyQt4'

profxj:
@marijana777 -- Make an inline comment under files changed
on where the offending line is.  Thanks.  Ask @jnburchett how 
to do it, if need be.

marijana777:
Solved. I needed to do git pull twice when switching the branch, @profxj

jessicawerk:
I would add all of these: 
ABMAGU          FLOAT         
   ABMAGG          FLOAT       
   ABMAGR          FLOAT          
   ABMAGI          FLOAT         
   ABMAGZ          FLOAT        
   ABMAGUERR       FLOAT          
   ABMAGGERR       FLOAT         
   ABMAGRERR       FLOAT        
   ABMAGIERR       FLOAT         
   ABMAGZERR       FLOAT       

For fun, you might add EBV, which I also put in there.

profxj:
ok, added all those and fixed EBV

@marijana777 -- this might be ready to merge but check.

marijana777:
- chalos_v12.cgm_abs[0].ebv has no value, and is not listed with vars. 
- Everything else works. 
- Colors u - r are not the same as in Table 1 , Werk et al. 2013. It might be useful if it is explained how the same colors could be obtained.

jessicawerk:
That's weird that ur colors are not the same as those published. The only thing I can vaguely remember has something to do with luptitudes (and negative fluxes) especially in the u-band. 
There is a column in the galaxy structure (megastruct) -- X -- called 
URCOLOR        
   URCOLOR_ERR    
I believe I dealt with the issue and added a tag for this very reason. 
May need to add those as well.

profxj:
Added as ur and ur_err

Merging.

PR #161, opened 2017-11-30
Add upper limit flag to dummy components

jnburchett:
As stated. Sets flag_N=3 to dummy components added during CgmAbsSys instantiation.


profxj:
merging

PR #163, opened 2017-12-08
CGM stack plots

jnburchett:
Create stack plots for CGM systems, accepting user-defined lists of ions to plot, number of transitions, etc.  Depends on linetools PR #436

### DIFF
--- a/pyigm/cgm/cgm.py
+++ b/pyigm/cgm/cgm.py
@@ -160,7 +160,7 @@ class CGMAbsSys(object):
         # Impact parameter and PA
         if rho is None:
             rho, iang = calc_cgm_rho(galaxy, igm_sys, self.cosmo, ang_sep=ang_sep,
-                                 correct_lowz=correct_lowz, **kwargs)
+                                     correct_lowz=correct_lowz, **kwargs)
             if debug:
                 pdb.set_trace()
             if ang_sep is None:
@@ -172,10 +172,10 @@ class CGMAbsSys(object):
         # Standard name
         if name is None:
             self.name = 'J{:s}{:s}_{:d}_{:d}'.format(
-                    self.igm_sys.coord.icrs.ra.to_string(unit=u.hour,sep='',pad=True)[0:4],
-                    self.igm_sys.coord.icrs.dec.to_string(sep='',pad=True,alwayssign=True)[0:5],
-                    int(np.round(self.PA.to('deg').value)),
-                    int(np.round(self.ang_sep.to('arcsec').value)))
+                self.igm_sys.coord.icrs.ra.to_string(unit=u.hour,sep='',pad=True)[0:4],
+                self.igm_sys.coord.icrs.dec.to_string(sep='',pad=True,alwayssign=True)[0:5],
+                int(np.round(self.PA.to('deg').value)),
+                int(np.round(self.ang_sep.to('arcsec').value)))
         else:
             self.name = name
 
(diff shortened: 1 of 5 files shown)


jnburchett:
I took a swing at the Travis failures...probably a miss.

pyigm/cgm/cgm.py
```diff
@@ -160,7 +160,7 @@ def __init__(self, galaxy, igm_sys, cosmo=None, name=None, rho=None, PA=None,
         # Impact parameter and PA
         if rho is None:
             rho, iang = calc_cgm_rho(galaxy, igm_sys, self.cosmo, ang_sep=ang_sep,
-                                 correct_lowz=correct_lowz, **kwargs)
+                                     correct_lowz=correct_lowz, **kwargs)
```
profxj: No tabs, please.

pyigm/cgm/cgm.py
```diff
@@ -260,3 +260,94 @@ def write_json(self, outfil=None):
         # Finish
         print("Wrote {:s} system to {:s} file".format(self.name, outfil))
 
+    def stack_plot(self,to_plot,pvlim=None,maxtrans=3,return_fig=True,**kwargs):
```
profxj, line 263: Is it not possible to link to the linetools method instead?
Or is there a good reason for this to be its own method here?

pyigm/utils.py
```diff
@@ -382,3 +382,32 @@ def velcorr_mould(galaxy,cosmo=None):
                        distmod=distmod, flag=flag, scale=scale.to(u.kpc/u.arcsec))
 
     return velcorrdict
+
+def get_components(obj,ion,zrange=None):
+    '''Return list of components from AbsSystem or IGMSightline for given ion
```
profxj: I believe the convention is to use """ instead of '''.

profxj (changes requested):
Am mainly concerned about having another method which at least partly duplicates what is in linetools.  

Explain that and we're good to go.

pyigm/cgm/cgm.py
jnburchett: I don't understand.  Could you elaborate?

jnburchett: Will move the correct_lowz... stuff back to the left in case that's what you're talking about.

pyigm/cgm/cgm.py
jnburchett: Let's say I want to programmatically generate a bunch of stack plots for CGM systems (as in a CGM survey).  Each call to generate the plot would also have to include the logic to extract a list of AbsLines to include in the plot.  In a real situation, I might just want to plot O VI and H I systems for each CGM, but the survey spans a large range in redshift where O VI may be covered at all redshifts but Ly-alpha may not.  Or, the whole Lyman series may be covered but I don't want to include every single transition in the stack plot.  Lines 291-339 deal with such issues and save the user from duplicating a bunch of complicated code simply for the sake of using the linetools function.

pyigm/utils.py
jnburchett: Changed.  Although I've probably used ''' in other places.  Will try and watch out for that.

jnburchett:
Please merge if you are happy with my rationale, @profxj. Thanks!

pyigm/cgm/cgm.py
profxj: Sounds good.

profxj:
merging.

PR #164, opened 2017-12-14
Survey component table

jnburchett:
Create table containing component information a la CGMSurvey.ion_tbl() but for components.

### DIFF
--- a/pyigm/cgm/cgmsurvey.py
+++ b/pyigm/cgm/cgmsurvey.py
@@ -227,7 +227,42 @@ class CGMAbsSurvey(object):
         return tbl
 
     def component_tbl(self, Zion, vlim=None):
-        return
+        """ Generate a Table of line measurements for an input ion broken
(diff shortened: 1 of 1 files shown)


pyigm/cgm/cgmsurvey.py
```diff
@@ -226,6 +226,43 @@ def ion_tbl(self, Zion, fill_ion=True):
         # Return
         return tbl
 
+    def component_tbl(self, Zion, vlim=None):
+        """ Generate a Table of line measurements for an input ion broken
+            down by component.
+
+            Parameters
```
profxj: Lines 233-243 should not be indented

profxj (commented):
make the minor change and feel free to merge.

pyigm/cgm/cgmsurvey.py
jnburchett: Good catch

PR #165, opened 2017-12-20
CGMAbsSys.igm_sys updates

jnburchett:
Addresses issues related to updating component data within a given CGM system's igm_sys where updates to various CGM systems associated with the same IGMSystem would propagate to all associated to CGM systems.


pyigm/cgm/utils.py
```diff
@@ -258,7 +259,13 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, rho_max=300*u.kpc, dv_max=400
         # Loop to generate
         cgm_list = []
         for imatch in match:
-            cgm = CGMAbsSys(galaxy, igmsystems[imatch], cosmo=cosmo, **kwargs)
+            # Instantiate new IGMSystem
+            # Otherwise, updates to the IGMSystem cross-pollinate other CGMs
+            sysmatch = igmsystems[imatch]
+            newisys = IGMSystem(sysmatch.coord,sysmatch.zabs,
```
profxj: have been advised to avoid copy statements when we can.
Can we?

profxj (changes requested):
See my pretty major comment.

profxj:
As a possible example, see the copy() method
of XSpectrum1D in linetools:

https://github.com/linetools/linetools/blob/master/linetools/spectra/xspectrum1d.py

pyigm/cgm/utils.py
jnburchett: Hmm...part of the issue was that updates were propagating through the AbsComponent objects too!  In fact, merely copying the IGMSystem didn't fix the problem, so the deepcopy was the best solution I could think of.  We could of course reinstantiate AbsComponents, but that's less than ideal for obvious reasons...suggestions?

profxj: I probably need to better understand the problem here.

If one does update the IGMSystem, wouldn't you want that
to propagate?  If not, I do feel you should re-instantiate.

jnburchett: Just to clarify, yes, one would want the main attributes of a component (column density, b-value, transitions) to propagate but the same IGM system will have different 'systemic' component velocities when associated to one galaxy (CGM systems) versus another.  The .copy methods are very useful anyway, so I'll go that direction.

jnburchett:
Okay, @profxj, think we should be good to go here.  Note that the changes depend on new linetools PR #445:  https://github.com/linetools/linetools/pull/445

profxj:
This all looks fine.  Merge when you wish.

jnburchett:
Great...will merge once the linetools dependency is clear.

profxj:
If we are clear @jnburchett , please merge.

jnburchett:
Weird issue occurring, and I wonder if it's an astropy bug.  Basically, SkyCoord objects with multiple coordinates fail when transforming, i.e., coords.icrs returns this:

*** TypeError: For this input type lists must contain either int or Ellipsis

Could you try this on your machine?  I was using astropy 2.0.2 and tried upgrading to 2.0.3 but get the same error.  This did not happen until we merged add_richter, but I don't think that's the issue.

PR #167, opened 2018-01-10
fixing igmgesses for new linetools

ntejos:
Fixing IGMGuesses for new linetools handling of b values. Backwards compatibility of old IGM_model.json files implemented.


profxj:
Looks fine.
merging.

PR #168, opened 2018-01-12
Several changes for --wotta option

cwotta:
Revamped MCMC_output() class
Updated saved info to include --wotta option info

### DIFF
--- a/pyigm/metallicity/mcmc.py
+++ b/pyigm/metallicity/mcmc.py
@@ -446,7 +446,8 @@ class Emceebones(object):
         else:
             lutext = "False"
         
-        plot_title = "{}\nUVB={}, log U prior={}".format(self.info['name'], self.UVB, lutext)
+        # plot_title = "{}\nUVB={}, log U prior={}".format(self.info['name'], self.UVB, lutext)
+        plot_title = "{}".format(self.info['name'])
 
         
         ######
--- a/pyigm/metallicity/mcmc.py
+++ b/pyigm/metallicity/mcmc.py
@@ -425,7 +425,34 @@ class Emceebones(object):
                     out_group[out_key] = self.final[out_key]
         except:
             pass
-                    
+        
+        ############
+        ##Plot
+        ############
+        
+        ######Setup
+        if self.ndim < 4:
+            title_xpos = 0.45
+            title_ypos = 0.80
+            title_fontsize = 12
+        else:
+            title_xpos = 0.35
+            title_ypos = 0.85
+            title_fontsize = 14
+        
+        
+        if self.logUconstraint.lower() == 'true':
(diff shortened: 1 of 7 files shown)


pyigm/metallicity/mcmc.py
```diff
@@ -459,17 +490,26 @@ def plotinfo(self, sampler, use_pkl=False):
...
-        #now do a corner plot
+
+        ######
+        ##Corner Plot
+        ######
+        
         samples = sampler.chain[:,self.burn:, :].reshape((-1,self.ndim))
-        cfig = corner.corner(samples, labels=self.mod_axistag, quantiles=[0.05,0.5,0.95],verbose=False)
-        cfig.text(0.33,0.90,self.info['name']+", UVB="+self.UVB+", logUconstraint="+str(self.logUconstraint),horizontalalignment='left',fontsize=10)
+        cfig = corner.corner(samples, labels=self.mod_axistag, label_kwargs = {"fontsize": 16}, quantiles=[0.05,0.5,0.95],verbose=False)
```
profxj, line 504: Is this for an updated corner method?

pyigm/scripts/pyigm_mtlmcmc.py
```diff
@@ -142,14 +174,16 @@ def run_mcmc_wotta(args):
     ##Test whether to include the carbalpha parameter, and use the correct UVB
     ##...But set a default first
     if not args.grid:
-        args.grid = 'grid_minimal'
+        args.grid = '/afs/crc.nd.edu/group/CGMND/Cloudy_grids/grid_cgm_extensive'
```
profxj: Let's not hard code this.  Can you have it be 
read in as a parameter?

profxj (commented):
Only 2 minor comments.
Address these and I'll merge.

profxj:
And thanks for this.  
I love all the documentation!

pyigm/metallicity/mcmc.py
cwotta: No, this is the same plot.  I'm just adding (1) a plot title (to match the other plots) and (2) adjusting font size of the axis titles (x and y labels).  I now have the plot title as a variable, too, so it's defined in a single place.

pyigm/scripts/pyigm_mtlmcmc.py
cwotta: It *is* read in as a parameter (see, e.g., docs/LLS_Metallicities-Wotta/queue_script.sh).  This line just sets a default if it doesn't receive an argument.

cwotta: (see also the documentation at the bottom of the script for input arguments)

profxj:
Merging.

PR #169, opened 2018-01-13
Refactor IGMSurvey (and children)

profxj:
Pretty significant Refactor of IGMSurvey.

Transtions away from a list of IGMSystem objects
towards an underlying Table or dict (or both).  

Positives:

- Much faster to load (and write)
- Cleaner access to data for most users
- Lightweight

Negatives:

- Might be confusing that the data can be stored in multiple places
- Is a step away from using the pyigm/linetools objects we spent all that time building


profxj:
Relies on PR#448 that is pending in _linetools_

profxj:
Ok @ntejos and @jnburchett 
This one is ready for review now.
Thanks for your diligence.

profxj:
Sorry, on second thought it is better to do this
PR from a later branch.  Closing this one.

PR #171, opened 2018-01-26
fix when self.quantity not in axis list

cwotta:
fix in get_walkers_*() for when self.quantity is not one of the axes.  simply return a blank list and nsys=0, like we do elsewhere.

update/correct docs at top


profxj:
Looks good to me.
Merging.

PR #174, opened 2018-02-10
Pulled UVB integration out; Python3 fix

cwotta:
Pulled function that integrates the UVB out of logU_to_dens converter function, allowing repeated called to logU_to_dens (or dens_to_logU) without reintegrating the spectrum every time.  Spectrum can now be passed as input to those converter functions.

Python 3 fix (xrange vs range)

"Initial guesses" file now allows the initial guesses to be False, if you'd like to set optim=True.  Useful for the first, quick, full-grid run for finding the maximum.  Then you'd follow up with an initial guess close to that maximum for a better result.  I was previously doing this, but it's much easier now.

Small documentation updates

### DIFF
--- a/docs/LLS_Metallicities-Wotta/queue_script.sh
+++ b/docs/LLS_Metallicities-Wotta/queue_script.sh
@@ -15,7 +15,7 @@
 basename "$0"
 ##Prints date+time to STDOUT
 ##  (to get an idea for how long this took to run)
-date
+date '+%Y-%m-%d %H:%M:%S'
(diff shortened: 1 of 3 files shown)


cwotta:
Just pinging @profxj on this.  Is it a problem that the Travis checks failed?  I didn't really change much.

profxj:
Yes, @cwotta I am worried about the tests, e.g.
https://travis-ci.org/pyigm/pyigm/builds/339889815

Can you pull the master branch into yours and see
if that fixes most of them?  I think it should.

PR #175, opened 2018-02-12
Remove astropy helpers

profxj:
And some other edits to get LLSSurvey passing tests.

And updated the LLSSurvey Notebook too.

Had to turn off coveralls but that's ok.

Passing all tests except the one plaguing astropy development.


profxj:
For what it is worth, @jnburchett , the numpy 1.14 failure
is there in 2.7 but not in 3.6:

https://travis-ci.org/pyigm/pyigm/builds/340378334

ntejos (approved):
Looks good to me.

profxj:
Merging.

PR #176, opened 2018-02-12
Igmguesses IO in IGMSightline

ntejos:
- Updates to IGMSightlines for IGMGuesses IO capabilities.
- Includes updates to IGMGuesses itself


profxj:
First, let's clear #175 as it helps with tests and
eliminates all of that astropy-helpers junk.

That said, I am reviewing this PR now.

pyigm/igm/igmsightline.py
```diff
@@ -39,7 +39,7 @@ def from_json(cls, jsonfile, **kwargs):
         return slf
 
     @classmethod
-    def from_igmguesses(cls, radec, zem, igmgfile, name=None, **kwargs):
+    def from_igmguesses_old(cls, radec, zem, igmgfile, name=None, **kwargs):
```
profxj: I think you can remove this method and have this
functionality in the new one by making radec and zem
optional keywords, e.g.

radec=None, zem=None

and then

``` 
if radec is not None:
         blah blah
```

pyigm/igm/igmsightline.py
```diff
@@ -133,10 +173,90 @@ def make_igmsystems(self, igmsystem=None, **kwargs):
...
+            out_dict['cmps'][key]['reliability'] = str(comp.reliability)
+            out_dict['cmps'][key]['comment'] = str(comp.comment)
+            # out_dict['cmps'][key]['mask_abslines'] = comp.mask_abslines
+
+        # JSONify
+        gd_dict = ltu.jsonify(out_dict)
+
+        # Write file
+        # with io.open(outfile, 'w', encoding='utf-8') as f:
+        f = open(outfile, 'w')
```
profxj: use  ltu.savejson() instead

profxj (approved):
just a few minor edits requested.

ntejos:
Ok, waiting now for PR #175 before attempting merging.

profxj:
PR #175 is cleared.  You might pull master into this
branch and make sure the tests all pass before merging.

ntejos:
Roger. Will see what the tests say...

ntejos:
Test failing (partially) because it depends on linetools' PR #458.

profxj:
Have cleared the linetools PR.  Check again
when you can.

ntejos:
Failing in numpy 1.14 when comparing to the written file with the reference file. In my computer there is no such an error, neither in the other versions of numpy. May be the sorting of the components once loaded and then written. Will check by modifying linetools to give a more meaningful message.

ntejos:
Failing some tests that are unrelated to this PR. 

Also failing 1 test in numpy 1.14 where the written file has a different number for 'wrest' than the reference. This is most likely coming from the fact that `AbsLines` inside the list `AbsComponent._abslines` are sorted in a different way depending on the numpy version. We could fix this by imposing that those abslines are sorted in a very specific way in linetools, but I think is beyond the scope of this PR. Would like to merge if you agree.

ntejos:
@profxj may we clear this as it is?

profxj:
Merging.

PR #177, opened 2018-02-14
Ricther+17 Nans

profxj:
Adds in the Ricther Notebook I had built but forgot to include.
And fixed some NANs that crept in to his Table..

Also added a modification to the igmsurvey.ions() method.
I hope it works ok with @jnburchett as I prefer to use the
masked Table instead of filling with null values.  Try
it out Joe!

### DIFF
--- a/pyigm/surveys/igmsurvey.py
+++ b/pyigm/surveys/igmsurvey.py
@@ -463,7 +463,7 @@ class IGMSurvey(object):
             raise ValueError("Not sure how to load the ions")
 
     # Get ions
-    def ions(self, Zion, Ej=0., skip_null=True):
+    def ions(self, Zion, Ej=0., skip_null=True, pad_with_nulls=False):
         """ Generate a Table of columns and so on
         Restrict to those systems where flg_clm > 0
 
@@ -474,7 +474,9 @@ class IGMSurvey(object):
         Ej : float [1/cm]
            Energy of the lower level (0. is resonance)
         skip_null : boolean (False)
-           Skip systems without an entry, else pad with zeros 
+           Skip systems without an entry, else pad with zeros
+        pad_with_nulls : bool, optional
+           Pad missing/null systems with empty values.  A bit risky
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


jnburchett:
@profxj , going through this now...no doubt it will break my code, as I have at least 1 critical routine that looks for table['log_N'] == 0 to determine where upper limits are needed.  However, the fallout of this breakage will be relatively minor compared to the changes we've already made.  I will consult you offline on how to adapt then we can merge.

pyigm/surveys/igmsurvey.py
```diff
@@ -474,7 +474,9 @@ def ions(self, Zion, Ej=0., skip_null=True):
         Ej : float [1/cm]
            Energy of the lower level (0. is resonance)
         skip_null : boolean (False)
-           Skip systems without an entry, else pad with zeros 
+           Skip systems without an entry, else pad with zeros
+        pad_with_nulls : bool, optional
+           Pad missing/null systems with empty values.  A bit risky
```
jnburchett, line 479: Why is this risky?   Perhaps it's due to my ignorance, but dealing with tables having null values is much less fussy than masked tables.

pyigm/surveys/igmsurvey.py
```diff
@@ -463,7 +463,7 @@ def fill_ions(self, use_Nfile=False, jfile=None, use_components=False,
             raise ValueError("Not sure how to load the ions")
 
     # Get ions
-    def ions(self, Zion, Ej=0., skip_null=True):
+    def ions(self, Zion, Ej=0., skip_null=True, pad_with_nulls=False):
```
jnburchett, line 466: Note, I updated CgmSurvey to pass , e.g., 'pad_with_nulls' through in calls to functions that require this method (like ion_tbl()).

jnburchett (approved):
I don't see what's gained here beyond complying to more elegant coding practices, as one is throwing away information (names, coordinates, redshifts) about systems when retrieving data.  However, I've pushed a commit to make this work with my existing CASBaH framework.  Merge if you like.

pyigm/surveys/igmsurvey.py
profxj: The main thing I was reacting against was hardcoding what the Table 
format will look like.  I can imagine it changing and then this will break.

profxj:
merging

PR #182, opened 2018-02-24
Pulled UVB integration to separate function

cwotta:
Biggest change: pull UVB integration out to separate function, so logU_to_dens() can be called repeatedly without re-integrating the spectrum each time.  Useful for converting a sightline's entire set of walkers from dens to logU.

Other changes: `--wotta` option now allows `optim=False`.  In `MCMC_initial_guesses-run_me.dat`, if either `met` or `dens` guess is set to False, `pyigm_mtlmcmc.py` will now set `optim=False` when sending to `mcmc.py`.

Minor changes: documentation

### DIFF
--- a/pyigm/metallicity/mcmc.py
+++ b/pyigm/metallicity/mcmc.py
@@ -61,7 +61,7 @@ from __future__ import print_function, absolute_import, division, unicode_litera
 import matplotlib as mpl
 
 #The next line works on queue systems for DISPLAY issues
-#mpl.use('Agg')   #  Commented out by X to avoid test failures
+mpl.use('Agg')
 
 import warnings
 import pdb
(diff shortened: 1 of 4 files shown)


pyigm/metallicity/mcmc.py
```diff
@@ -61,7 +61,7 @@
 import matplotlib as mpl
 
 #The next line works on queue systems for DISPLAY issues
-#mpl.use('Agg')   #  Commented out by X to avoid test failures
+mpl.use('Agg')
```
profxj: This may cause test failures..
If it is required then we might need
to change the tests..

profxj (commented):
Only one comment for you.

profxj:
Merging.

PR #183, opened 2018-03-08
CGM covering fractions, etc.

jnburchett:
Util for calculating covering fractions in a CGM survey as well as a general util (plus wrapper) for confidence intervals on integer ratios

### DIFF
--- a/pyigm/cgm/utils.py
+++ b/pyigm/cgm/utils.py
@@ -274,4 +274,64 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, rho_max=300*u.kpc, dv_max=400
     # Return
     return cgm_list
 
+def covering_fraction(iontable,colthresh,rhobins=[0,150],returncounts=False,**kwargs):
+    """Given some detection threshold, calculate the ion covering fraction
+    within impact parameter bins
 
+    Parameters
+    ----------
+    iontable : Table
+        Output of CGMAbsSurvey.ion_tbl()
+    colthresh : float or None
+        Detection threshold in log column density; if None, impose no threshold
+    rhobins,optional : list of ints or floats
+        Bins in impact parameter
+    returncounts : bool, optional
+        If True, return numbers of hits and total systems within each bin
+
+    Returns
+    -------
+    fracs : list of floats
+        Covering fraction for each bin
+    lolims : list of floats
+        Lower limits on covering fractions
+    uplims : list of floats
+        Upper limits on covering fractions
+    dethist : list of ints, optional
+        Number of detections in each bin
+    tothist : list of ints, optional
+        Total number of systems in each bin
+
+    """
+    from pyigm import utils as pu
+
+    # Clean up table to remove nulls if they exist
+    itab = iontable.copy()
+    itab = itab[itab['flag_N']>0]
+    flag = itab['flag_N']
+    col = itab['logN']
+    rho = itab['rho_impact']
+
+    # Impose threshold if it exists and
+    if colthresh != None:
+        dets = np.where(((flag == 1) | (flag == 2))&(col > colthresh))[0]
+        nondets = np.where((flag == 3) & (col <= colthresh))[0]
+    else:
+        dets = np.where((flag == 1) | (flag == 2))[0]
+        nondets = np.where((flag == 3))[0]
+
+    # Set up and sort bins
+    dethist, bins = np.histogram(rho[dets], bins=rhobins)
+    nondethist, bins = np.histogram(rho[nondets], bins=rhobins)
+
+    # Get the stats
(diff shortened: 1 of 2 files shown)


pyigm/cgm/utils.py
```diff
@@ -274,4 +274,64 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, rho_max=300*u.kpc, dv_max=400
...
 
+def covering_fraction(iontable,colthresh,rhobins=[0,150],returncounts=False,**kwargs):
+    """Given some detection threshold, calculate the ion covering fraction
+    within impact parameter bins
 
+    Parameters
+    ----------
+    iontable : Table
+        Output of CGMAbsSurvey.ion_tbl()
+    colthresh : float or None
```
profxj, line 285: Can we alternatively have an EW threshold?

pyigm/cgm/utils.py
```diff
@@ -274,4 +274,64 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, rho_max=300*u.kpc, dv_max=400
...
+        Detection threshold in log column density; if None, impose no threshold
+    rhobins,optional : list of ints or floats
+        Bins in impact parameter
```
profxj: If nothing is input, what is done?  Do you just get one 
number for the whole lot?

pyigm/cgm/utils.py
```diff
@@ -274,4 +274,64 @@ def cgm_from_galaxy_igmsystems(galaxy, igmsystems, rho_max=300*u.kpc, dv_max=400
...
+    itab = iontable.copy()
+    itab = itab[itab['flag_N']>0]
+    flag = itab['flag_N']
+    col = itab['logN']
+    rho = itab['rho_impact']
+
+    # Impose threshold if it exists and
+    if colthresh != None:
+        dets = np.where(((flag == 1) | (flag == 2))&(col > colthresh))[0]
+        nondets = np.where((flag == 3) & (col <= colthresh))[0]
```
profxj: We should comment above that non-detections above the 
threshold are simply ignored (which is sensible).

pyigm/utils.py
```diff
@@ -411,3 +411,60 @@ def get_components(obj,ion,zrange=None):
...
+def confintervals(hits,totals,sig=1):
+    """Wrapper for confinterval() function to accept arrays of hits and totals
+    
+    Parameters
+    ----------
+    hits : list or array
+        Numbers of instances
+    totals : list or array
+        Total numbers of instances possible
+    sig : int
```
profxj: Does it need to be an integer?

pyigm/utils.py
```diff
@@ -411,3 +411,60 @@ def get_components(obj,ion,zrange=None):
...
+
+def confinterval(hits,total,sig=1):
+    """ Calculate Wilson confidence intervals
+    Parameters
+    ----------
+    hits : int
+        Number of instances
+    total : int
+        Total number of instances possible
+    sig : int
```
profxj: same query as above.  It is possible someone
will want a 95% c.l. or a 99% c.l. and therefore
would insert a float.

profxj: Ah, I see below that you are translating to conf interval anyhow.
I might then suggest that one can provide either (and none defaults
to 1 sigma equivalent).   We'd then need to use a method to translate
sigma to conf which I am sure I have somewhere..  Either in pyigm or
linetools..

profxj (changes requested):
Only change I wish to impose is on inputing
sigma/conf-level in the utils method you generated.

pyigm/cgm/utils.py
jnburchett: This currently operates on the ion tables, which don't have EW.  Indeed, the table contains summed column densities system by system, and I'm not sure EW really is meaningful in the 'absorption system' context..

pyigm/cgm/utils.py
jnburchett: This defaulted to [0,150] kpc, but I like your suggestion (if it was one) better...changed.  I also clarified the units.

pyigm/cgm/utils.py
jnburchett: Done

pyigm/utils.py
jnburchett: Done!  I couldn't find a preexisting method, so I added a simple one.

jnburchett:
All good, @profxj ?

pyigm/cgm/utils.py
profxj: ah, right.  There is a method to generate 
a transition Table and one might wish to 
feed that in sometime, but it can wait 
another day.

profxj (approved):
looks great.
merge it

PR #185, opened 2018-04-07
Refactor CGM survey

profxj:
Primarily a refactoring of the CGMAbsSurvey object to instantiate
from a JSON file (or dict) without generating all of the objects.  This
speeds up instantiation considerably.  

Introduces _dict and _data and several methods to enhance _data

A few changes related to IGMSightline snuck in for CASBaH building.

This PR depends a little on PR #471 pending in linetools.

Several new tests.
No new docs.

PR #186, opened 2018-04-12
namespace for argparse

ntejos:
Namespace issue in argparse


ntejos:
Test failing when comparing two IGMGuesses files. There is a new key called "Id" that is causing problem.

profxj:
I've also touched the IGMGuesses file in the PR #185 

I guess I'll deal with that conflict after merging this..

PR #188, opened 2018-04-30
Covering fraction detection threshold

jnburchett:
Very simple fix to covering fraction calculations.  Code previously did not count detections below threshold as nondetections.


profxj:
merging.

PR #190, opened 2018-05-07
Clustering: Separate sensitivity function from random galaxies

profxj:
As titled.

A few bug fixes too and a bit more docs.

### DIFF
--- a/pyigm/clustering/clustering_field.py
+++ b/pyigm/clustering/clustering_field.py
@@ -103,12 +103,12 @@ class ClusteringField(IgmGalaxyField):
         # Load me up
         if self.galreal is None:
             self.galreal = galnew  # np rec array with galaxy properties
-            self.galrand = random_gal(self.galreal, self.Ngal_rand)
+            self.galrand, _, _, _ = random_gal(self.galreal, self.Ngal_rand)
         else:
             galnew = galnew.astype(self.galreal.dtype)
             self.galreal = np.append(self.galreal, galnew)
             self.galreal = np.rec.array(self.galreal)
-            aux = random_gal(galnew, self.Ngal_rand)
+            aux, _, _, _ = random_gal(galnew, self.Ngal_rand)
             self.galrand = np.append(self.galrand, aux)
             self.galrand = np.rec.array(self.galrand)
 
--- a/pyigm/clustering/clustering_field.py
+++ b/pyigm/clustering/clustering_field.py
@@ -89,8 +89,9 @@ class ClusteringField(IgmGalaxyField):
         z_clm : str, optional
           Name of the column for the galaxy redshifts
         sens_galaxies : np.recarray
-          Use for generating sensitivity function instead of the added galaxies
-          Used when the subset is too small for an accurate sensitivity function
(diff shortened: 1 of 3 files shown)


pyigm/clustering/clustering_field.py
```diff
@@ -63,13 +65,17 @@ def __init__(self, radec, R=20000, Ngal_rand=10, Nabs_rand=100, proper=False, fi
...
+            self.name = field_name
         self.proper = proper
         self.wrapped = wrapped
 
         self.galreal = None # Filled with addGal
 
-    def addGal(self, gal_idx, mag_clm='MAG', z_clm='ZGAL'):
+    def addGal(self, gal_idx, mag_clm='MAG', z_clm='ZGAL', sens_galaxies=None,
+               magbins=None, SPL=None):
         """ Adds a set of galaxies for clustering analysis from the main astropy Table
```
ntejos, line 79: Is it astropy Table still?

pyigm/clustering/clustering_field.py
```diff
@@ -82,6 +88,12 @@ def addGal(self, gal_idx, mag_clm='MAG', z_clm='ZGAL'):
           Name of the column for the galaxy magnitudes
         z_clm : str, optional
           Name of the column for the galaxy redshifts
+        sens_galaxies : np.recarray
+          Use for generating sensitivity function instead of the added galaxies
```
ntejos: Describe better... array containing the shape of the sensitivity function of the galaxies being added. Actually, I think it may be better to always require this sensitivity function to be provided by the user. What do you think?

pyigm/clustering/xcorr_utils.py
```diff
@@ -257,42 +259,39 @@ def random_gal(galreal, Nrand, Nmin=20, DZ=0.01, smooth_scale=10.,
...
     delta_mag : float, optional
        Step size for magnitude binning
+    delta_mag : float (optional)
+    debug : bool (optional)
 
     Returns
     -------
-    galrand : np rec array
-      Copy of galreal, Nrand times
+    VALS : dict
```
ntejos, line 267: Describe VALS: the actual histograms used to get the Cubic Spline (I think)

ntejos (commented):
Looks good to me, please merge but would be the right time to consider better coding to increase speed.

pyigm/clustering/clustering_field.py
profxj: self.galaxies is an astropy Table but it is not
involved in any of the clustering calculations.

pyigm/clustering/clustering_field.py
profxj: I found it helpful to have it automatically
generated for awhile.  But I agree it is 
a bit dangerous.

pyigm/clustering/xcorr_utils.py
profxj: added

profxj:
As for speeding up the code, it is rather fast for me (takes a few
seconds per field).  But I've not used more than 10 randoms thus far..

Will investigate speed when it becomes an issue.

profxj:
merging

PR #192, opened 2018-06-04
Adding missing ions in CGM stackplots

jnburchett:
Previously, it was necessary to have AbsLines of a given species already associated for CGMAbsSys.stack_plot() to produce a panel showing the spectral regions covering that species' transitions.  This adds that functionality.

### DIFF
--- a/pyigm/cgm/cgm.py
+++ b/pyigm/cgm/cgm.py
@@ -285,7 +285,8 @@ class CGMAbsSys(object):
         # Finish
         print("Wrote {:s} system to {:s} file".format(self.name, outfil))
 
-    def stack_plot(self,to_plot,pvlim=None,maxtrans=3,return_fig=True,**kwargs):
+    def stack_plot(self,to_plot,pvlim=None,maxtrans=3,return_fig=True,
+                   add_missing_lines=False,spec=None,**kwargs):
         '''Show a stack plot of the CGM absorption system
 
         Parameters
@@ -299,6 +300,11 @@ class CGMAbsSys(object):
            Override system vlim for plotting
         maxtrans : int, optional
            Maximum number of lines per transition to plot
+        add_missing_lines : bool, optional
+           If True, plot transitions that do not have associated AbsLine objects
+        spec : XSpectrum1D, optional
+           Spectrum to plot in regions of requested lines; required if assoc.
+           AbsLine objects do not have their analy['specfile'] attributes set
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


pyigm/cgm/cgm.py
```diff
@@ -323,10 +330,26 @@ def stack_plot(self,to_plot,pvlim=None,maxtrans=3,return_fig=True,**kwargs):
                 else: # Pick components in system closest to z_cgm
                     thesecomps = pu.get_components(self,tp)
                     if len(thesecomps)==0:
-                        continue
-                    # Find component with redshift closest to systemic
-                    compvels = np.array([np.median(tc.vlim.value) for tc in thesecomps])
-                    comp = thesecomps[np.argmin(np.abs(compvels))]
+                        if add_missing_lines is True:
+                            if isinstance(tp,str):
```
profxj: I'm not sure str is 2.7 compatible, but
that's why we are on the way to 3.X

pyigm/cgm/cgm.py
```diff
@@ -323,10 +330,26 @@ def stack_plot(self,to_plot,pvlim=None,maxtrans=3,return_fig=True,**kwargs):
...
+                                tup = ltai.name_to_ion(tp)
+                            else:
+                                tup = tp
+                            comp = AbsComponent(self.coord,tup,zcomp=self.z,
+                                                vlim=[-100.,100.]*u.km/u.s)
+                            comp.add_abslines_from_linelist()
+                            if spec is not None:
```
profxj, line 342: Maybe add a comment here.  I can't remember why the
first element is enough for setting ```spec```

pyigm/cgm/cgm.py
```diff
@@ -335,35 +358,53 @@ def stack_plot(self,to_plot,pvlim=None,maxtrans=3,return_fig=True,**kwargs):
...
-                        complines = np.array(complines) # For the indexing
-                        try:
-                            tokeep = [complines[compnames == sn][0] for sn in strong['name']]
-                            lines2plot.extend(tokeep)
-                        except:
-                            warnings.warn('{} covered by spectra but not in'
-                                          'components list'.format(sn))
+                        for i,sn in enumerate(strong['name']):
+                            if sn in compnames:
+                                tokeep = [complines[compnames == sn][0]]
```
profxj: Add a few comments in here if you can.
Thanks!

profxj (commented):
A few minor requests.
Merge when you are happy.

pyigm/cgm/cgm.py
jnburchett: It actually works under 2.7 on my desktop machine!

pyigm/cgm/cgm.py
jnburchett: Good catch!  It isn't enough.

pyigm/cgm/cgm.py
jnburchett: Done.

jnburchett:
merging

PR #194, opened 2018-06-13
NH data for COS-Halos

profxj:
From Prochaska+17

Gives instructions on how to download the MCMC traces
and then includes code to load them into the COSHalos
Class.


marijana777:
In https://github.com/pyigm/pyigm/blob/hi_cos_halos/docs/datasets/COS_Halos_Examples.ipynb I am getting different results in the last ~2 cells:

cgm_abs =  cos_halos.cgm_abs[0]
cgm_abs.igm_sys.NH_PDF     
--> <GenericPDF: mean=19.9336>    instead of  <GenericPDF: mean=19.9342>

cgm_abs.igm_sys.NH_PDF.gmean   
 --> 2.3340520914386813e+20    instead of 2.3504263052011087e+20

profxj:
Those differences are minor.

Are you ok with the rest of the PR, @marijana777 ?

marijana777:
Yes, I am OK with the rest, @profxj

PR #195, opened 2018-07-05
small change in LLS and COS-Dwarfs

marijana777:
1) Script pyigm_fitlls does not work in all computers, possibly because of python 2 vs python 3 differences. Debugging. 
2) Adding an option to define separate path to COS Dwarfs data (cdir).

### DIFF
--- a/pyigm/abssys/lls.py
+++ b/pyigm/abssys/lls.py
@@ -498,6 +498,7 @@ def tau_multi_lls(wave, all_lls, **kwargs):
     from linetools.analysis import voigt as lav
     #
     all_tau_model = np.zeros(len(wave))
+    all_tau_model = list(np.zeros(len(wave)))  ## need list not array for python3 ?
     # Loop on LLS
     for lls in all_lls:
         # LL
--- a/docs/fitlls.rst
+++ b/docs/fitlls.rst
@@ -63,8 +63,8 @@ Usage
 
 When the LLS gui launches, you should begin by left clicking
 in the plot window.  This will activate plotting and key strokes
-(see below for the help message).   You should then refine the
-continuum ("C", "1", "2").
+(see below for the help message). Then press "#" to specify the model spec and to add a continuum.
+You should then refine the continuum ("C", "1", "2").
 
 From there, navigate to the LLS of interest, add a model ("A" or "F"),
 refine it and the continuum, etc.
--- a/pyigm/cgm/cos_halos.py
+++ b/pyigm/cgm/cos_halos.py
@@ -858,6 +858,8 @@ class COSDwarfs(COSHalos):
         if cdir is None:
             self.cdir = os.environ.get('COSHALOS_DATA')
             #self.cdir = os.environ.get('DROPBOX_DIR')+'/COS-Dwarfs/'
+        else:
+            self.cdir = cdir  ### need to define cdir
         if fits_path is None:
             self.fits_path = os.path.abspath(os.environ.get('DROPBOX_DIR')+'/COS-Dwarfs/Targets/FITS')
         else:


pyigm/abssys/lls.py
```diff
@@ -498,6 +498,7 @@ def tau_multi_lls(wave, all_lls, **kwargs):
     from linetools.analysis import voigt as lav
     #
     all_tau_model = np.zeros(len(wave))
+    all_tau_model = list(np.zeros(len(wave)))  ## need list not array for python3 ?
```
profxj, line 501: Please explain why this needs to be a list.

profxj:
When you do a PR, a description is expected.  Please provide one.

pyigm/abssys/lls.py
marijana777: If this is not a list, it gives an error AttributeError: 'numpy.ndarray' object has no attribute '_set_unit', and the gui does not fit a LLS. With list it fits a LLS. I am not sure about exact reasons, but I am assuming that this is probably a difference between python 2 and 3.

profxj: There is no way that 

```
all_tau_model = np.zeros(len(wave))
```

will give a _set_unit error.

You must be referring to other code.  
Please expplain.

profxj (changes requested):

pyigm/abssys/lls.py
marijana777: The part of the code that is giving that error is
(line 519 in lls.py)   all_tau_model += tau_model

profxj: Then why make the change here?  Might it not break
things elsewhere?

marijana777: I would propose making the change in the same place, and adding at the end of the tau_multi_lls function something that would convert tau_multi_lls to its original type (analogue to all_tau_model = np.asarray(all_tau_model)). Could you let me know what is the original array type? Thanks.

marijana777: I tried adding all_tau_model = np.asarray(all_tau_model) at the end, but the gui was then not able to fit LLS. Also redshift for continuum fitting could not be QSO redshift. I suggest we discuss this more tomorrow, and if you can show me an example how is the gui supposed to work. I would also suggest adding docs on this gui.

profxj: Please do add docs on the GUI.  I'm making an Issue
and assigning it to you.

profxj:
@marijana777 -- Is this PR ready for merging now?

marijana777:
(2) Adding an option to define separate path to COS Dwarfs data (cdir). is ready for merging,
(1) could be ignored if it is solved in another way. Checking

marijana777:
Well...... The last version in master branch that I had was 6f769a3  (6/14/2018). I did git pull, and tried pyigm_fitlls again, and got error  
Traceback (most recent call last):
  File "/home/smailagic/anaconda3/bin/pyigm_fitlls", line 11, in <module>
    load_entry_point('pyigm', 'console_scripts', 'pyigm_fitlls')()
  File "/home/smailagic/pyigm/pyigm/scripts/pyigm_fitlls.py", line 54, in main
    lls_fit_file=pargs.lls_fit_file, norm=args.norm)
AttributeError: 'NoneType' object has no attribute 'norm'

The version in llsdw branch is at least working reasonably well.

marijana777:
Is this PR ready for merging now? - - I would say yes.

jnburchett:
@marijana777 are you waiting on your changes to be accepted? I see one outstanding review from @profxj, but I'm not sure if this is the hold up.  Thanks!

marijana777:
@jnburchett , yes, I am waiting for my changes to be accepted, or for @profxj to let me know if he wants me to change something more. Thanks for asking!

profxj (approved):
merging

PR #204, opened 2018-09-21
Generalized covering fraction

jnburchett:
Enables one to calculate covering fractions based on any property in the supplied table

Also, modifies cgm association algorithm to hunt for components closer to galaxy
redshift than abssystem z


profxj (approved):
Looks good.

profxj:
merging

PR #206, opened 2018-12-13
fitdla and fitlls guis - adding options (see below)

marijana777:
1) making arguments masking and norm optional
2) adding option conti_file='from_spectrum' to use continuum data from the original spectrum

### DIFF
--- a/docs/fitdla.rst
+++ b/docs/fitdla.rst
@@ -40,6 +40,7 @@ script `pyigm_fitdla`.  Here is the usage::
                             Input DLA Fit file
       --conti_file CONTI_FILE
                             Input continuum spectrum
+                            Option 'from_spectrum': continuum data will be used from the original (in_file) spectrum.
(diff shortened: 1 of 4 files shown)


docs/fitdla.rst
```diff
@@ -40,6 +40,7 @@ script `pyigm_fitdla`.  Here is the usage::
                             Input DLA Fit file
       --conti_file CONTI_FILE
                             Input continuum spectrum
+                            Option 'from_spectrum': continuum data will be used from the original (in_file) spectrum.
```
profxj, line 43: Add that 100 spline points will be added uniformly 
along the input continuum.

profxj (approved):
One minor request.

docs/fitdla.rst
marijana777: Added.

profxj:
merging

PR #207, opened 2019-01-17
Igmg

ntejos:
Minor change to IGUGuesses, adding a function to create an AbsComponent list from an IGMGuesses .json output file.

[not sure where did the other commits came from]


profxj (approved):
all looks good to me.

profxj:
merging.

PR #209, opened 2019-02-21
Igmg: Allowing to grab transitions from LineList H2

ntejos:
Eventually, more LineLists can be included in the check for transitions.


profxj (approved):
A bit of a kludge, but that is fine for `igmguesses`

profxj:
merging.

ntejos:
Indeed. It would be nice to have a proper way to deal with H2 absorption within linetools... eventually.

PR #210, opened 2019-02-22
Adding option to smooth/unsmooth spectrum in Igmguesses

marijana777:
* - smooth spectrum, / - unsmooth spectrum

### DIFF
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -127,6 +127,7 @@ y         : guess y limits
 t,b       : set y top/bottom limit
 l,r       : set left/right x limit
 [,]       : pan left/right
+*,/       : smooth/unsmooth spectrum
 C,c       : add/remove column
 K,k       : add/remove row
 (         : toggle between many/few (15 or 6) panels per page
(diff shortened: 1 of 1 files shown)


pyigm/guis/igmguesses.py
```diff
@@ -616,6 +617,7 @@ def __init__(self, ispec, z, parent=None, llist=None, norm=True, screen_scale=1.
         self.parent = parent
         self.spec = ispec
         self.spec_fil = self.spec.filename
+        self.orig_spec = ispec  # For smoothing
```
profxj: You may need to use `.copy()` here

profxj (commented):
Consider the comments

pyigm/guis/igmguesses.py
profxj: And `.copy()` when you re-assign `self.spec` for un-smoothing.

ntejos:
Thanks @marijana777 for this. How about using the same keystrokes for smoothing/unsmoothing as in the linetools' GUIs? e.g lt_xspec?

pyigm/guis/igmguesses.py
marijana777: Thanks @profxj . Now using .copy(), in both cases. (Also, continuum-normalized orig_spec, and defined bad_pixels.)

marijana777:
Thanks @ntejos . I think that this would be a good idea. 
However, U and S are already used in different way in igmguesses (U - restrict to subset of available lines from the parent LineList at current redshift, S - select an absorption component from cursor position)

profxj:
merging.

PR #211, opened 2019-02-25
adding documentation for H2 LineList.

ntejos:
If we want a different keystroke, this is a good time to request it.


profxj (approved):
approving

profxj:
merging

PR #221, opened 2019-08-22
Fix Richter

profxj:
Mainly a PR to fix velocity issues with the Richter+17 dataset.

But, I've cleaned up a few tests along the way.


profxj:
merging

PR #223, opened 2019-08-23
Halo model edits

profxj:
Strips out the remainder of the halo models.
These are now in the FRB repository:  https://github.com/FRBs/FRB

But does add in a few files need for halo calculations,
e.g. Faerman+17, Stern+18.  These may eventually migrate to FRB..


profxj:
merging

PR #224, opened 2019-09-17
Adding utility function is_not_HILya()

ntejos:
As stated. A simple function that identifies pixels that belong to a metal line or a higher-order HI Lyman absorption from an input wavelength array. Useful for masking out interlopers.

### DIFF
--- a/pyigm/surveys/utils.py
+++ b/pyigm/surveys/utils.py
@@ -7,6 +7,7 @@ import glob
 import json
 import pdb
 
+import numpy as np
 from astropy.coordinates import SkyCoord
 
 from linetools import utils as ltu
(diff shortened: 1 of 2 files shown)


pyigm/surveys/utils.py
```diff
@@ -119,3 +120,33 @@ def class_by_type(type):
...
+    Parameters
+    ----------
+    wobs : Quantity array
+        Observed wavelength
+    comp : list of AbsComponent objects
+        List of abscomponent objects identified
+
+    Returns
+    -------
+    answer : bool
```
profxj: this is a boolean array, not just a bool, right?
If so, kindly change this doc string

profxj (approved):
One minor thing to address.
Am approving.

pyigm/surveys/utils.py
ntejos: Documentation improved.

jnburchett (approved):

profxj:
Merging

PR #225, opened 2019-12-30
Adding an option to igmguesses to calculate transitions over a wider wavelength range than the input spectrum

marijana777:
In this way, two spectra of the same object with different wavelength range could share the same .json igmguesses output file and transitions in that file.

### DIFF
--- a/pyigm/guis/igmguesses.py
+++ b/pyigm/guis/igmguesses.py
@@ -85,7 +85,8 @@ class IGMGuessesGui(QMainWindow):
     def __init__(self, ispec, parent=None, previous_file=None, 
         srch_id=True, outfil=None, fwhm=None, screen_scale=1.,
         plot_residuals=True,n_max_tuple=None, min_strength=None,
-                 min_ew=None, vlim_disp=None, external_model=None, redsh=None): #added redsh=None
+                 min_ew=None, vlim_disp=None, external_model=None, redsh=None,
+                 spwvmin=None,spwvmax=None): #added redsh=None
         QMainWindow.__init__(self, parent)
         """
         ispec : str
@@ -206,11 +207,24 @@ E         : toggle displaying/hiding the external absorption model
             vlim_disp = [-500.,500.] * u.km/u.s
         self.vlim_disp = vlim_disp
 
+        #####################
+        #self.wvmin = 1383. *u.AA
+        #self.wvmax = 3691. *u.AA
 
 
         # Load spectrum
         # spec, spec_fil = ltgu.read_spec(ispec, masking='edges')
         spec = readspec(ispec, masking='edges')
+
+        if spwvmin is not None:
+            self.wvmin = spwvmin *u.AA
+        else:
+            self.wvmin = np.min(spec.wavelength)
+        #
+        if spwvmax is not None:
+            self.wvmax = spwvmax *u.AA
+        else:
+            self.wvmax = np.max(spec.wavelength)
 
 
         # Normalize
@@ -467,6 +481,7 @@ E         : toggle displaying/hiding the external absorption model
                 # QtCore.pyqtRemoveInputHook()
                 # pdb.set_trace()
                 # QtCore.pyqtRestoreInputHook()
+                #pdb.set_trace()
                 comp_init_attrib(comp)
                 comp.init_wrest = igmg_dict['cmps'][key]['wrest']*u.AA
                 try:
@@ -474,6 +489,7 @@ E         : toggle displaying/hiding the external absorption model
                 except KeyError:  # For compatibility
                     warnings.warn("Setting all abslines to 2")
                     comp.mask_abslines = 2*np.ones(len(comp._abslines)).astype(int)
+                #pdb.set_trace()
                 self.velplot_widg.add_component(comp, update_model=False)
 
             else:  # for compatibility, should be deprecated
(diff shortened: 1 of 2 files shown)


pyigm/guis/igmguesses.py
```diff
@@ -742,6 +767,13 @@ def update_model(self):
             return
         # Setup lines
         wvmin, wvmax = np.min(self.spec.wavelength), np.max(self.spec.wavelength)
+        spwvmin = self.parent.wvmin
+        spwvmax = self.parent.wvmax
+        if spwvmin is not None:
```
profxj: I don't think `spwvmin` can ever be `None`.
Isn't it set by default to `min(spec.wavelength)` ?

If so, is this doing what you want it to do?

pyigm/guis/igmguesses.py
```diff
@@ -778,19 +810,25 @@ def add_component(self, inp, vlim=None, zcomp=None, update_model=True):
...
+
         if isinstance(inp, AbsComponent):
 
             # get rid of lines outside spectral coverage, if any
             # Todo: one may want to reconsider this, specially if more spectral coverage is obtained
             new_abslines = []
             for absline in inp._abslines:
                 wobs = absline.wrest * (1 + absline.z)
-                if (wobs > self.spec.wvmin) and (wobs < self.spec.wvmax):
+                if (wobs > self.parent.wvmin) and (wobs < self.parent.wvmax):
```
profxj, line 819: Should you be calling the parent `wvmin` or 
one in this object?  Seems to me the latter,
so please double check.

pyigm/guis/igmguesses.py
```diff
@@ -827,7 +865,8 @@ def add_component(self, inp, vlim=None, zcomp=None, update_model=True):
                 linelist_aux = self.llist[current_parent_linelist]
 
             # add the component
-            new_comp = create_component(zcomp, inp, linelist_aux, vlim=vlim, spec=self.spec)
+            new_comp = create_component(zcomp, inp, linelist_aux, vlim=vlim, spec=self.spec,
+                                        spwvmin=self.parent.wvmin, spwvmax=self.parent.wvmax)
```
profxj, line 867: Same as above.  Why are we still referencing the parent `wvmin` ?

pyigm/guis/igmguesses.py
```diff
@@ -1858,7 +1901,7 @@ def create_component(z, wrest, linelist, vlim=[-300.,300]*u.km/u.s,
 
 def comp_init_attrib(comp):
     # Attributes
-    comp.attrib = {'N': 0./u.cm**2, 'Nsig': 0./u.cm**2, 'flagN': 0,  # Column
+    comp.attrib = {'N': 0./u.cm**2, 'Nsig': 0./u.cm**2, 'flag_N': 0,  # Column
```
profxj: While it is `flagN` in other places, it isn't in `igmguesses`.
Are you sure this won't break things?  @ntejos -- please comment as well.

profxj (changes requested):
Have a few questions to be addressed before approving.

profxj:
too stale

pyigm/guis/igmguesses.py
marijana777: I agree that spwvmin could not be None. Yes, the code was doing what I wanted it to do, but there was no need to write if spwvmin is not None, because it will always be true. Correcting (deleted if spwvmin is not None, and similar).

pyigm/guis/igmguesses.py
marijana777: Checked again. When igmguesses is used with -p guesses.json -o guesses.json option, calling self.spec.wvmin and self.spec.wvmax deletes absorption lines from guesses file that are not in the spectral range of the input spectrum (i.e., within [self.spec.wvmin, self.spec.wvmax]). On the other hand, calling self.parent.wvmin and self.parent.wvmax includes all absorption lines in a customly defined wavelength range [wmin,wmax]. In this way, one guesses file could be used when working with multiple non-overlapping spectra (or spectra with different wavelength ranges) of the same object. Therefore, it should be self.parent.wvmin and self.parent.wvmax.

pyigm/guis/igmguesses.py
marijana777: Checked again. For the same or similar reason as above. When igmguesses is used with -p guesses.json -o guesses.json option, calling self.spec.wvmin and self.spec.wvmax does the following: When a new absorption component is defined, the absorption component contains only absorption lines within the wavelength range of a spectrum. (i.e., within [self.spec.wvmin, self.spec.wvmax]). On the other hand, calling self.parent.wvmin and self.parent.wvmax includes all absorption lines in a customly defined wavelength range [wmin,wmax]. In this way, one guesses file could be used when working with multiple non-overlapping spectra (or spectra with different wavelength ranges) of the same object. Therefore, it should be self.parent.wvmin and self.parent.wvmax.

pyigm/guis/igmguesses.py
marijana777: Igmguesses works with either option. I remember that I was getting an error (maybe when using another gui?), and that I solved it by changing flagN to flag_N. But, now I could not remember more details. I agree to keep flagN at this moment.

marijana777:
Responded to the questions and made small changes related to (1) and (4). So I would reopen the PR.

PR #228, opened 2020-12-28
Documentation regarding a (uncommon) bug error in IGMGuesses

ntejos:
Extra documentation regarding an uncommon bug in IGMGuesses.


profxj (approved):
good good.  merge

